PR #3, opened 2018-09-30
Refactor local post usage

vujevits:
#### todos
- [x] LocalPostManager removed
- [x] local posts are in redux state and persisted
- [ ] loading a subset of the posts from storage
- [ ] add integration tests to master and merge it to this branch
- [ ] reimplement DebugScreen


src/reducers/index.ts
```diff
@@ -109,9 +132,23 @@ const draftReducer = (draft: Post | null = null, action: Actions): Post | null =
     return draft;
 };
 
+const metadataReducer = (metadata: Metadata = { highestSeenId: 0 }, action: Actions): Metadata => {
+    switch (action.type) {
+        case 'UPDATE-HIGHEST-SEEN-ID': {
+            return {
+                ...metadata,
+                highestSeenId: metadata.highestSeenId++,
```
agazso: `metadata.highestSeenId + 1` would be nicer here, because the `++` has also an unintended side effect.

vujevits: 👍

PR #8, opened 2018-10-06
Create user

vujevits:
- design is rudimentary
- favicons are persisted for each post, I believe as intended


src/components/IdentitySettings.tsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const Header = (props: { onPressBack: () => void }): React.ReactElement<{}> => {
+    return (
+        <View style={styles.headerContainer}>
+            <View style={{ paddingLeft: 15 }}>
+                <TouchableOpacity
+                    onPress={props.onPressBack}
+                    activeOpacity={1.0}
+                    hitSlop={{
```
agazso: You can also use TouchableView here which has a default hitSlop of 30

src/reducers/index.ts
```diff
@@ -35,12 +36,19 @@ const defaultSettings: Settings = {
     saveToCameraRoll: true,
 };
 
+const defaultAuthor: Author = {
+    name: '',
+    uri: '',
```
agazso: Where is uri used and what is it supposed to mean?

src/reducers/index.ts
```diff
@@ -35,12 +36,19 @@ const defaultSettings: Settings = {
...
+    faviconUri: '',
```
agazso: Why is this called faviconUri? I think imageUri or avatarUri would be a better name.

src/components/IdentitySettings.tsx
```diff
@@ -0,0 +1,118 @@
...
+                        top: 10,
+                        left: 10,
+                        bottom: 10,
+                        right: 10,
+                    }}
+                >
+                <Text style={{ color: '#007AFF' }}>
```
agazso: Can you make this a bit bigger? :)

src/components/IdentitySettings.tsx
vujevits: fixed in 9be93f9328558afb5e61d224446f772a29cb0e32

src/components/IdentitySettings.tsx
vujevits: fixed in 9be93f9328558afb5e61d224446f772a29cb0e32

PR #9, opened 2018-10-14
kind of nice solution for introducing welcome screen

vujevits:


src/App.tsx
```diff
@@ -144,12 +144,33 @@ const AppNavigator = StackNavigator(Scenes,
...
+        screen: IdentitySettingsContainer,
+    },
+}, {
+    mode: 'card',
+    navigationOptions: {
+        header: null,
+    },
+});
+
+const SwNavigator = SwitchNavigator({
```
agazso: I think MainNavigator or InitialNavigator would be a better name for this

agazso:
👍

agazso:
One thing I found: there is a Back button on the Welcome screen, it would be great to remove it and put a title instead.

vujevits:
Introduced a 2 page onboarding where on the second screen the username and the avatar can be selected

src/App.tsx
vujevits: solved

PR #31, opened 2018-10-28
adding default posts to localposts

vujevits:
3 default posts in the yourfeed, one of them contains a gif showing how to add an RSS.
#### Issues
- ugly hack to load it in EditFeed, which could be solved by a refactor of our ImageData type - instead of uri, we should store ImagePropertiesSourceOptions
- Contrary to what I said earlier, the posts are persisted, so they 'can be' / 'have to be' deleted, which is good, but there is a warning because of new posts having the same id


agazso:
To your second point: can't you simply initialize the id as well in the defaultState?

PR #33, opened 2018-11-04
Added random keypair generation for base identity

agazso:


src/reducers/index.ts
```diff
@@ -46,12 +46,7 @@ const defaultAuthor: Author = {
     image: {
         uri: '',
     },
-    identity: {
-        publicKey: '',
-        privateKey: '0x0bb80d052b3d85656b3f082c24fb6c211d7935163fab6c605358886e56d17a93',
-        // address: '0xADc9b12480cE9880D6Bed1Ef91Cdc279D671Cf0d',
-        address: '0x377820b4e6913dd322a24e2af10e1c51c5f1e709',
-    },
+    identity: undefined,
```
vujevits: not a big deal, but this line can be omitted

agazso: I wanted to make it explicit, so that we know it's not forgotten, but it's undefined by design.

PR #80, opened 2018-12-11
remove onboarding posts when additional posts are added

### LINKED ISSUE #70: In the onboarding screen after editing the name, you have to tap twice to open the photo selector

vujevits:
closes #70

### DIFF
--- a/src/reducers/index.ts
+++ b/src/reducers/index.ts
@@ -1,4 +1,4 @@
-import { List } from 'immutable';
+import { List, Iterable } from 'immutable';
 import {
     createStore,
     combineReducers,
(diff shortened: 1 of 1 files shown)


src/reducers/index.ts
```diff
@@ -247,9 +248,24 @@ const rssPostsReducer = (rssPosts = List<Post>(), action: Actions): List<Post> =
...
+        return list.remove(ind);
+    } else {
+        return list;
+    }
+};
+
 const localPostsReducer = (localPosts = defaultLocalPosts, action: Actions): List<Post> => {
     switch (action.type) {
         case 'ADD-POST': {
+            if (localPosts.findIndex(post => post != null && post._id === defaultPost1._id) >= 0) {
```
agazso: I see a potential performance problem here. Anytime you add a post, this will iterate through all the posts minimum once. This also means that it will get much worse after you have many posts.
In theory it would be enough to trigger the condition when the action.payload.post._id  equals defaultLocalPosts.size, no?

PR #81, opened 2018-12-12
Open feed

### LINKED ISSUE #62: open individual feed

vujevits:
Opens a feed in a new view. Works with RSS feeds as expected. The local / not uploaded may need some more fine-tuning.
closes #62


src/actions/Actions.ts
```diff
@@ -49,6 +50,8 @@ const InternalActions = {
         createAction(ActionTypes.INCREASE_HIGHEST_SEEN_POST_ID),
     addOwnFeed: (feed: PostFeed) =>
         createAction(ActionTypes.ADD_OWN_FEED, { feed }),
+    openFeed: (feedUrl: string) =>
```
agazso: Is this used at all?

src/components/YourFeed.tsx
```diff
@@ -243,13 +245,16 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
         const url = post.link || '';
         const hostnameText = url === '' ? '' : ' -  ' + Utils.getHumanHostname(url);
         return (
-            <View style={styles.infoContainer}>
+            <TouchableOpacity
+                onPress={() => this.props.navigation.navigate('Feed', { feedUrl: post.author && post.author.uri })}
```
agazso: I think it would be better to pass `post.author` here and then we could display the name of the feed as the title of the navigation header.

src/components/YourFeed.tsx
```diff
@@ -350,11 +355,21 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
...
+        if (this.props.displayFeedHeader) {
+            return (
+                <FeedHeader
+                    navigation={this.props.navigation}
+                    onSavePost={this.props.onSavePost}
+                />
+            );
+        } else {
+            return (
+                <NavigationHeader
```
agazso: Just an idea: we could put an Unfollow button on the right side so that you don't have to go to the settings and find a feed to unfollow.

agazso:
I think it would be better if the tabbar and the navigation header stayed visible when you navigate to a feed. I know the tabbar is a bit more complicated, because of how React Navigation works, but at least the NavigationHeader should stay always visible so that it's easy to go back (at least on iOS where there is no back button otherwise).

src/actions/Actions.ts
vujevits: no, I decided to go with the infamous react-navigation params. It was late and there is some leftover code 😿

src/components/YourFeed.tsx
vujevits: good idea

src/components/YourFeed.tsx
vujevits: same as above :)

vujevits:
I believe I implemented everyting we have talked about, there is one remaining issue: the tabs are stateful this way, so opening a feed -> change tab -> go back to previous tab -> shows the opened feed instead of news/yourfeed

PR #90, opened 2018-12-20
Favorite

vujevits:
Because how I changed the UI I decided to add a 'refollow' feature which escalated a bit :-)
I added to the state the visitedFeeds. Originally I wanted something for storing unfollowed feeds, but I realized we will want to open feeds without following them explicitly, hence the naming. 
One thing is stil missing (if I'm not mistaken), removing the posts from unfollowed feed, when navigating away.


src/actions/Actions.ts
```diff
@@ -18,6 +18,7 @@ export enum ActionTypes {
     CLEANUP_CONTENT_FILTERS = 'CLEANUP-CONTENT-FILTERS',
     ADD_FEED = 'ADD-FEED',
     REMOVE_FEED = 'REMOVE-FEED',
+    TOGGLE_FEED_FAVORITE = 'TOGGLE-FEED-FAVORITE',
```
agazso: This action is not used in any reducers, so it seems to be not working.

src/containers/FavoritesContainer.ts
```diff
@@ -0,0 +1,62 @@
...
+    return feeds.find(feed => {
+        return feed != null && post.author != null &&
+            feed.feedUrl === post.author.uri &&
+            feed.favorite === true;
+    }) != null;
+};
+
+const mapStateToProps = (state: AppState, ownProps): StateProps => {
+    const posts = state.rssPosts
+        .filter(post => post != null && isPostFromFavoriteFeed(post, state.feeds))
```
agazso: It would be better to create a variable that holds the favorite feeds and then pass that list to isPostFromFavorites.

vujevits:
@agazso Added some changes we have discussed, please try how it works for you now. There are some code quality issues left, imho does should be addressed outside this pr.

src/components/YourFeed.tsx
```diff
@@ -42,15 +44,19 @@ export interface DispatchProps {
...
 export interface StateProps {
     navigation: any;
     posts: Post[];
+    feeds: Feed[];
+    knownFeeds: Feed[];
     settings: Settings;
-    displayFeedHeader: boolean;
-    showUnfollow: boolean;
+    yourFeedVariant: YourFeedVariant;
+    notOwnFeed: boolean;
```
agazso: Why is the naming inverse? Usually it's easier to follow the logic if the variable name is direct, e.g. `isOwnFeed`

src/components/YourFeed.tsx
```diff
@@ -151,6 +154,53 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
...
+                        /> : undefined}
+                        rightButtonText2={props.notOwnFeed ? <MaterialIcon
+                            name={'favorite'}
+                            size={20}
+                            color={isFollowedFeed
+                                ? props.isFavorite() ? Colors.BRAND_RED : Colors.DARK_GRAY
+                                : 'transparent'
+                            }
+                        /> : undefined}
+                        onPressRightButton1={async () => props.notOwnFeed && await props.onFollowPressed(props.navigation.state.params.author)}
```
agazso: Why is this async? The signature of `onFollowPressed` does not tell that it's async.

src/components/YourFeed.tsx
```diff
@@ -151,6 +154,53 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
...
+                        onPressRightButton2={() => props.notOwnFeed && isFollowedFeed && props.onToggleFavorite(props.navigation.state.params.author.uri)}
+                        title={props.navigation.state.params ? props.navigation.state.params.author.name : 'Favorites'}
```
agazso: It would be better to use a variable for `author` and not use the whole `props.navigation.state.params.author...` everywhere

src/components/YourFeed.tsx
```diff
@@ -164,11 +214,18 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
...
     }
 
-    private async unfollowFeed(author: Author) {
-        const confirmUnfollow = await AreYouSureDialog.show('Are you sure you want to unfollow?');
-        if (confirmUnfollow) {
-            this.props.onUnfollowFeed(author.uri);
-            this.props.navigation.goBack(null);
+    private onFollowPressed = async (author: Author) => {
+        const followedFeed = this.props.feeds.find(feed => feed.feedUrl === author.uri);
+        if (followedFeed != null) {
```
agazso: I don't really understand what's going on here: sometimes if the follow pressed we unfollow a feed, sometimes we add the feed to the feeds list.
In the unfollow case I think the naming is confusing and maybe the UI as well.
In the other knownFeed case, it seems we add the knownFeed to the feed list one more time, because the only case when the `onAddFeed` does anything is in the `FeedContainer`, where `knownFeeds` is just a filtered list of `state.feeds`.

src/components/YourFeed.tsx
vujevits: you cannot put the async keyword there, but it's on the function what we are calling.

src/components/YourFeed.tsx
vujevits: 👍

src/components/YourFeed.tsx
```diff
@@ -217,15 +218,23 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
...
+        const confirmUnfollow = await AreYouSureDialog.show('Are you sure you want to unfollow?');
+        if (confirmUnfollow) {
+            this.props.onUnfollowFeed(feed);
+        }
+    }
+
+    private followFeed(author: Author) {
+        const knownFeed = this.props.knownFeeds.find(feed => feed.feedUrl === author.uri);
+        if (knownFeed != null) {
+            this.props.onAddFeed(knownFeed);
```
agazso: The logic is still incorrect here: `knownFeeds` is coming from `state.feeds`, so if you find a known feed, you will add it to `state.feeds` therefore it will be added twice.

vujevits: it is not, check the reducer. Maybe the callback should be renamed but not sure to what

src/reducers/index.ts
```diff
@@ -169,13 +169,20 @@ const contentFiltersReducer = (contentFilters = List<ContentFilter>(), action: A
 const feedsReducer = (feeds = List<Feed>(defaultFeeds), action: Actions): List<Feed> => {
     switch (action.type) {
         case 'ADD-FEED': {
```
agazso: Ok, so I understand what's going on here. `ADD-FEED` doesn't just add a feed, it also involves some logic, if the feed is already added then it will update the existing feed to be followed.
In that case `ADD-FOLLOWED-FEED` or something similar would be a better name, because by convention all the existing ADD-* actions are just adding to a list. Also the name of the function should be changed, because this is very confusing.

PR #98, opened 2018-12-29
Navigation header

agazso:
This is an attempt to fix https://github.com/agazso/postmodern/issues/57

- Got rid of react-navigation navigation header everywhere
- Changed the styling to match the platform better (arrows, colors, right button placement)
- The NavigationHeader can be also used as the list header, which is now the default behavior for the news, favorite and generic feeds. Maybe it would be better to create a generic Header component and a NavigationHeader with default back behavior?

I prefer this to the original, because it looks much better (especially on android) and can be easily styled and more flexible. However we lost the iOS header transition animations with this, but maybe we can implement something similar with react-native. Also on iPhoneX* the header is 'jumping' when switching between the tabs, which is very annoying.


vujevits:
LGTM

Issue #87, opened 2018-12-18
Remove async functionality from YourFeed


agazso:
Sometimes I see unfinished Promise errors coming from YourFeed. It may come from legacy/unused code doing offline network detection, updating the elapsed time in posts etc.

PR #99, opened 2018-12-30
Refactor your feed

vujevits:
[WIP]
- needs more manual testing
- some tests
- merge https://github.com/agazso/postmodern/pull/90 first


src/components/Card.tsx
```diff
@@ -0,0 +1,333 @@
...
+    togglePostSelection: (post: Post) => void;
+    openPost: (post: Post) => void;
+    onDeleteConfirmation: (post: Post) => void;
+    onSharePost: (post: Post) => void;
+    navigate: (view: string, {}) => void;
+}
+
+export const Card = (props: CardProps) => {
+    if (props.post.images.length === 0) {
+        return <CardWithText {...props}/>;
```
agazso: This is not necessary, the else branch renders the same when there are no images and without CardWithText there is no code duplication.

src/components/Card.tsx
```diff
@@ -0,0 +1,333 @@
...
+                        : <ActionIcon name={shareIconName}/>
+                    }
+                </TouchableView>
+            </View>
+        );
+    }
+    return <View/>;
+};
+
+const renderCardTopIcon = (post: Post) => {
```
agazso: This could be a component too

src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,22 @@
...
+export class ErrorBoundary extends React.Component<{}, { hasError: boolean }> {
+    public static getDerivedStateFromError(error) {
+        // Update state so the next render will show the fallback UI.
+        return { hasError: true };
+    }
+
+    public state = { hasError: false };
+
+    public componentDidCatch(error, info) {
+        console.log(error, info);
```
agazso: We should use Debug.log everywhere

src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public render() {
+        if (this.state.hasError) {
+            return <Text>Something went wrong.</Text>;
```
agazso: It's better to not show error to the user when there are no meaningful action that they can do, so I would just return with null here (and log the error of course)

test/components/CardTest.tsx
```diff
@@ -0,0 +1,175 @@
+import * as React from 'react';
+import * as ShallowRenderer from 'react-test-renderer/shallow';
+import { Card, CardWithText } from '../../src/components/Card';
+import { Author, Post } from '../../src/models/Post';
+import TestRenderer from 'react-test-renderer';
+
+jest.mock('react-native-fs', () => {
```
agazso: It seems easier to mock `getLocalPath` as we discussed :)

src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import { Text } from 'react-native';
+
+export class ErrorBoundary extends React.Component<{}, { hasError: boolean }> {
```
agazso: Cool, I didn't know this is possible 👍

src/components/YourFeed.tsx
```diff
@@ -224,33 +202,33 @@ export class YourFeed extends React.PureComponent<DispatchProps & StateProps, Yo
...
             if (!isSwarmLink(post.link)) {
                 await Linking.openURL(post.link);
             }
         }
     }
 
-    private isPostSelected(post: Post) {
-        return this.state.selectedPost && this.state.selectedPost._id === post._id;
+    private isPostSelected = (post: Post): boolean => {
+        return !!this.state.selectedPost && this.state.selectedPost._id === post._id;
```
agazso: Is this a new convention that we are using instead of `!= null` now?
The old code wasn't conforming either, but this is new to me too.

src/components/ErrorBoundary.tsx
vujevits: I have a hunch this is not true, let's compare the 2 screenshots.

src/components/Card.tsx
vujevits: There are some minor differences, but I just moved the code :-)

src/components/Card.tsx
vujevits: 👍

src/components/YourFeed.tsx
vujevits: no, this was accidental here. The only time when I intent to use it is with optional boolean fields.

src/components/ErrorBoundary.tsx
agazso: What I'm saying is that when you return this text, the user will see a post with the text 'Something went wrong'. From a user perspective this is not so good, because this is not the original text of the post, but some users may think that is. Other users may think that there was an error with this post, but since there are no actions that they can do about it, it's not so useful.
I think the best would be to hide posts that has errors in it (as well as not let users post content that we cannot display later), but I don't know how to do it at this stage.

vujevits: Also depends on what we mean by logging. If we don't want automatic / opt-in logging with some service, it might be useful to give the users (maybe only in the beta or a tester user group) some kind of error message with an actionable item - sending the logs. I've opened an issue for it https://github.com/agazso/postmodern/issues/103

agazso:
This is related to #87 and #89

PR #108, opened 2019-01-04
Redux migration for versioned store

agazso:
This has some issues right now:
- Because we are using immutable-js version 3.X, the Typescript typings are not correct and for example the argument of the function passed to `map()` is always optional, and that makes it hard to write simple expressions to replace `List`s. Instead we have to use `toArray()` and then convert back the whole thing to immutable `List`. This bug will be fixed in immutable-js version 4.X, but currently it's at RC version for almost a year now, so the development is not too active.
- It's not completely type-safe, because it casts the state parameter to the AppState which has already the new version we want to convert to. It's not trivial to make it completely type-safe, because that would require us to version all the different types as well and that does not seem sustainable.


src/reducers/migration.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Debug } from '../Debug';
+import { AppState } from '.';
+import { Feed } from '../models/Feed';
+
+const migrateUnversionedToVersion0 = (state: PersistedState): AppState => {
+    Debug.log('Migrate unversioned to version 0');
+    const appState = state as AppState;
+    const version0AppState = {
+        ...appState,
+        feeds: List<Feed>(appState.feeds.toArray().map(feed => ({
```
vujevits: I've read the description in the PR, but don't really get what the issue is. Is there an actual bug - we cannot use map on Immutable.List, or is it just an error in the d.ts file? If the latter, maybe we should monkey patch it in a post install script.

agazso: Unfortunately there are other problems with the type definitions and the immutable library itself. `map` returns an `Iterable` type that needs to be converted to a list (with `toList()`) but then the types are again wrong.

I will leave this as is for now.

Issue #106, opened 2019-01-04
Backup and restore data


agazso:
Idea: store the data encrypted with password on swarm with a link that can be shared.
Problems: Swarm encryption can only be used securely with light node, but not with gateway, so this has to be done after light node is integrated.

Idea: use a One Time Pad to encrypt the data, maybe with using additional encryption.
Problems: we will two links, one to the actual encrypted data and an other one to the OTP. Also this requires to store twice the size of the actual data on Swarm.

vujevits:
What data do you mean here? Your not shared private/personal, - at the moment offline - posts? If we are planning to upload everything, isn't it simpler to have a specific feed which only you have access to, and then the only thing you have to backup, is your private key.

PR #111, opened 2019-01-10
Backup & restore

agazso:
Basic backup & restore functionality, hidden in the debug menu at the moment.

The idea is to use the serialized state that is stored in LocalStorage then:
- encode it in UTF-8
- encrypt it with NaCL (XSalsa20+Poly1305) with the SHA3 of the user-provided password
- then encode it in base64
- finally wrap it with a header and footer

The resulting data can be shared with the OS provided apps. To restore the backup do the inverse of each step in a reversed order. 

Ideas to make it better:
- Use the privateKey (or something equivalent like mnemonics) as a password when encrypting. This has the advantage, that the user is advised to somehow guard this information, but the UX of remembering 256 bit information is a bit cumbersome. I experimented with mnemonics, and it's possible to store the entropy that is required to generate the privateKey in 192 bits, but that's still 18 english words to remember in the correct order.
- Store data on Swarm, make the whole backup process automatic. For this we should not trust the users to provide a password, because then their whole data can become accessible for unauthorised parties if they brute force or guess the weak password. At this moment I think the best would be to use a One Time Pad encryption with an additional backup password. See the discussion here: https://github.com/agazso/postmodern/issues/106


src/BackupRestore.ts
```diff
@@ -0,0 +1,56 @@
...
+    const encryptedData = nacl.secret_box.formatWN.pack(dataUint8Array, secureRandomUint8Array, secretUint8Array);
+    const encryptedDataBase64 = base64.encode(encryptedData);
+    const backupText = wrapWithHeaderAndFooter(encryptedDataBase64);
+    return backupText;
+};
+
+export const restoreBackupToString = async (backupText: string, secretHex: string): Promise<string> => {
+    const encryptedDataBase64 = stripHeaderAndFooter(backupText);
+    const secretUint8Array = new Uint8Array(keccak256.array(hexToByteArray(secretHex)));
+    const encryptedBytes = new Uint8Array(base64.decode(encryptedDataBase64));
```
vujevits: shouldn't base64.decode come after stripHeaderAndFooter?

agazso: No, the header and footer is non-binary data that is useful to provide human readable information about the backup. For example we provide a version field that the app can use to decide if it is able to open the backup or not. With this it's possible to change for example the encoding to base85 or something else in a later version and it will be easy to handle that change from the code.

I used a similar model that PGP uses for encrypting emails, see https://pitt.edu/%7Epoole/PGP.htm#step8

vujevits: that's not what I meant, that is clear. I'm sure you tried it and works, but seemingly the order is not the reverse of the other function (the penultimate step is encode, so here the second step should be decode)

agazso: The second step here is to process to passed in hex secret string for passing it to the encryption and decryption function, which is the same in both cases. It is not strictly part of the encryption process, however it's necessary to do it to get the code work. This could be potentially the zeroth step.

vujevits: oh I see it now, 👍

PR #126, opened 2019-01-13
Refactor your feed 2

### LINKED ISSUE #89: YourFeed is having too many responsabilities

vujevits:
fixes #89 by using children prop to inject view dependent UI. This simplifies the code a lot, also the empty functions code smell from the containers are gone. The only logic now in the old YourFeed - RefreshableFeed atm, is the refresh, and the post selection. Another dependency is the Card, if it doesn't cause issues in testing I would leave it as it is, we can always move it out later if we want to have a feed of other then Posts.

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -33,55 +33,51 @@ export interface DispatchProps {
 type CardProps = StateProps & DispatchProps;
 
 export const Card = (props: CardProps) => {
-    if (props.post.images.length === 0) {
-        return <CardWithText {...props}/>;
-    } else {
-        return (
-            <View
-                style={[styles.container, {
-                    margin: 0,
-                    padding: 0,
-                    paddingTop: 5,
-                    borderWidth: 0,
-                }]}
-                key={'card-' + props.post._id}
-                testID={'YourFeed/Post' + props.post._id}
-            >
+    return (
+        <View
+            style={[styles.container, {
+                margin: 0,
+                padding: 0,
+                paddingTop: 5,
+                borderWidth: 0,
+            }]}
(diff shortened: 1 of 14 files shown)


src/components/Card.tsx
```diff
@@ -32,7 +32,7 @@ export interface DispatchProps {
 
 type CardProps = StateProps & DispatchProps;
 
-export const Card = React.memo((props: CardProps) => {
+export const Card = (props: CardProps) => {
     if (props.post.images.length === 0) {
```
agazso: This `if` is not necessary anymore, both branches render the same thing now.

src/components/Card.tsx
```diff
@@ -82,7 +82,9 @@ export const Card = React.memo((props: CardProps) => {
             </View>
         );
     }
-});
+};
+
+export const MemoizedCard = React.memo(Card);
```
agazso: Where do we use the plain (not memoized) `Card`? If we don't use it, can't we just use the name `Card` for the memoized version?

src/components/RefreshableFeed.tsx
```diff
@@ -0,0 +1,138 @@
...
+export interface DispatchProps {
+    onRefreshPosts: (feeds: Feed[]) => void;
+}
+
+export interface StateProps {
+    navigation: any;
+    posts: Post[];
+    feeds: Feed[];
+    settings: Settings;
+    children: {
```
agazso: Why are these passed in as children and not top-level properties?

src/components/Card.tsx
vujevits: ok

PR #127, opened 2019-01-14
Feed preview

### LINKED ISSUE #94: after fetching a feed, open it in the 'Feed view', so the user can decide to follow it or not

vujevits:
fixes #94 
Also added a button to open a feed from the former 'EditFeed'
Merge #126 first

### DIFF
--- a/src/components/FeedInfo.tsx
+++ b/src/components/FeedInfo.tsx
@@ -79,7 +79,7 @@ export class FeedInfo extends React.Component<DispatchProps & StateProps, EditFe
             });
             this.onAdd(feed);
             await this.props.onDownloadFeed(feed);
-            this.props.navigation.navigate('Feed', { uri: feed.feedUrl, name: feed.name });
+            this.props.navigation.navigate('Feed', { feedUrl: feed.feedUrl, name: feed.name });
         } else {
             this.onFailedFeedLoad();
         }
@@ -112,7 +112,7 @@ export class FeedInfo extends React.Component<DispatchProps & StateProps, EditFe
         ;
         const rightButtonAction2 = this.state.loading || !isExistingFeed
             ? undefined
-            : async () => await this.props.navigation.navigate('Feed', { uri: this.props.feed.feedUrl, name: this.props.feed.name })
+            : async () => await this.props.navigation.navigate('Feed', { feedUrl: this.props.feed.feedUrl, name: this.props.feed.name })
         ;
         return (
             <View style={styles.container}>
--- a/src/components/FeedInfo.tsx
+++ b/src/components/FeedInfo.tsx
@@ -1,9 +1,7 @@
 import * as React from 'react';
 import {
-    TextInput,
     Alert,
     StyleSheet,
-    Button,
     View,
     Text,
     ActivityIndicator,
@@ -23,6 +21,7 @@ import * as Swarm from '../Swarm';
 import { downloadPostFeed } from '../PostFeed';
 import { NavigationHeader } from './NavigationHeader';
 import Icon from 'react-native-vector-icons/MaterialCommunityIcons';
+import { unfollowFeed } from './FeedView';
 
 const QRCodeWidth = Dimensions.get('window').width * 0.6;
 const QRCodeHeight = QRCodeWidth;
@@ -38,6 +37,7 @@ interface EditFeedState {
 export interface DispatchProps {
     onAddFeed: (feed: Feed) => void;
     onRemoveFeed: (feed: Feed) => void;
+    onUnfollowFeed: (feed: Feed) => void;
 }
 
 export interface StateProps {
@@ -85,10 +85,15 @@ export class FeedInfo extends React.Component<DispatchProps & StateProps, EditFe
 
     public render() {
         const isExistingFeed = this.props.feed.feedUrl.length > 0;
+        const isFollowed = this.props.feed.followed;
         const rightButtonText1 = this.state.loading
             ? undefined
             : <Icon
-                  name={isExistingFeed ? 'delete' : 'download'}
+                  name={isExistingFeed
+                    ? isFollowed
+                        ? 'link-variant-off'
+                        : 'delete'
+                    : 'download'}
                   size={20}
                   color={Colors.DARK_GRAY}
               />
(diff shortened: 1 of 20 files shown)


src/components/NewsFeedView.tsx
```diff
@@ -0,0 +1,34 @@
...
+                listHeader: <NavigationHeader
+                                leftButtonText=''
+                                title='News'
+                            />,
+            }}
+        </RefreshableFeed>
+    );
+};
+
+export const MemoizedNewsFeedView = React.memo(NewsFeedView);
```
agazso: As far as I understand, right now we don't gain anything with memoization, because we use `immutable-js` and we generate the props from it any time the state changes. 

One solution could be to get rid of `immutable-js`. Recently I tend to like this idea, because there is not a lot we gain by using it, but we have several problems with it (like this one, or the incorrect type defs).

src/components/FeedView.tsx
```diff
@@ -0,0 +1,94 @@
...
+export interface DispatchProps {
+    onRefreshPosts: (feeds: Feed[]) => void;
+    onFollowFeed: (feed: Feed) => void;
+    onUnfollowFeed: (feed: Feed) => void;
+    onToggleFavorite: (feedUrl: string) => void;
+}
+
+export interface StateProps {
+    navigation: any;
+    uri: string;
```
agazso: `uri` is not really a good name, because of two reasons:
- It does not tell what does it belong to (I guess it's the feed uri, but I am not sure)
- we should not use `uri` anywhere in the code, because it's meaning is too broad. I know it's already used at other places, but we should change those as well. The best would be to use a different type for this, but then we should use a smarter JSON serializing/deserializing library than the built-in ones.

Instead we can just use url everywhere, in this case `feedUrl`.

src/components/FeedInfo.tsx
```diff
@@ -76,36 +77,55 @@ export class EditFeed extends React.Component<DispatchProps & StateProps, EditFe
                 checked: true,
                 loading: false,
             });
-            await this.onAdd(feed);
+            this.onAdd(feed);
+            await this.props.onDownloadFeed(feed);
```
agazso: This is not necessary, because the onAddFeed implementation in the container also triggers the download.

If you want to keep it for some reason, `await` here is not necessary, because it calls the onDownloadFeed which is not async.

src/components/FeedInfo.tsx
```diff
@@ -37,14 +38,15 @@ interface EditFeedState {
 export interface DispatchProps {
     onAddFeed: (feed: Feed) => void;
     onRemoveFeed: (feed: Feed) => void;
+    onDownloadFeed: (feed: Feed) => void;
```
agazso: This is not necessary, because the onAddFeed implementation in the container also triggers the download.

agazso:
I think the UI flow would be better if it worked like this:

For viewing a feed:
- when you tap an item in the Feed list, it opens the feed view, with a favorite button and a button for the info view instead of the unfollow
- then you can open the feed info, and there you can unfollow (delete if already unfollowed) and favorite (if followed)
- going back would go back to the feed view, then to the feed list

For adding a new feed:
- it opens the feed info screen and you can enter the link or QR code
- then if it can open the feed, it shows you the feed contents, with the unfollow and favorite buttons. This means that the feed is automatically added to the followed feeds.
- Going back goes to the feed list

This way the feed list could be reused on other screens. I was thinking of putting a button to the right side of the Favorite and News tab navigation headers, which would be a dropdown list of the feeds that are present in the respective tabs. Later we could add a search option to that as well and that could be part of the solution the 'feeds are hard to find' problem.

src/components/NewsFeedView.tsx
agazso: Also because the serialization of immutable Lists and Maps are stupidly inefficient (it doubles the size of the backup compared to plain JSON).

src/components/FeedView.tsx
vujevits: good point

vujevits: I renamed it in the navParams as well

src/components/FeedInfo.tsx
vujevits: Good catch!
about await, I did not look into it just followed what you did :-)
Which is await when the dispatched action is async, are you saying that does not work like that?

vujevits:
- I agree with back should go back to the feed list, that is the planned behavior, it's just not done yet
- the reason why I tend to agree with the 'auto-follow' behavior, is that the current icon is not good enough, maybe we should try addressing that first?
- I think reusability can be solved by moving the navigation function to the dispatchprop, and having a different container for other usecases

src/components/FeedInfo.tsx
agazso: Here I just stated that the `onDownloadFeed` function is by definition is not an async function in the `DispatchProps` so no `await` is needed.

It doesn't matter if the dispatched action is async or not from the perspective of the function that is calling the `dispatch`, because `dispatch` is not an async function. I don't know how it's implemented exactly, but I suspect that the async actions are put in a queue and executed later.

vujevits:
here is the gist of the latest commit:
- unfollow or delete button depending on current feed state
- both actions navigate back to feedlisteditor
- opening feed and going back takes back to the feedinfo

There is an issue which had to be worked around: passing something as a navParam, which originates from the redux-store, introduces state inconsistency, if you let's say navigate BACK to a screen, which previously received something as a navParam. This can be solved by calling the setParam method on the navigation. 
The other case when it causes issues is not when navigating back, but directly modyfing it when you are on the screen with the navParam. For this I decided to just navigate back to the FeedList, getting rid of the screen in the navigation history with the invalid state.

@agazso please try the functionality

agazso:
When adding a new feed, they are not followed by default, so they are greyed out. I thought we agreed on that right now adding a new feed will follow it automatically. I think the current behavior is very confusing.

Because of this, when adding a new feed, the favorite button is missing as well.

Also trying this out I think after adding a feed the back button should go back to the feed list. Also when opening a feed from the feedlist should show the feed, with an info button on the top, that takes you to the feed info screen (but this may be a separate PR). I think this because you rarely need the feed info screen itself (only when deleting or sharing with someone).

src/components/NewsFeedView.tsx
agazso: I would keep the memoized version, because after getting rid of immutable it will be better. Also I would change the names so that NewsFeedView is the memoized version and there would be a NotMemoizedNewsFeed (or some better name but you get the point :)

vujevits: the classic discussion about IServiceProvider + ServiceProvider OR ServiceProvider + DefaultServiceProvider/ServiceProviderImpl...
I have no opinion, so be it :-)

vujevits:
The issues were dealt with, let's look at later on the last suggestion.

PR #129, opened 2019-01-17
Avatar caching

agazso:


android/app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,7 @@
     android:versionName="1.0">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```
vujevits: I assume this is a natural requirement for applications, but why is it necessary now?

src/FileSystem.ts
```diff
@@ -0,0 +1,20 @@
+import RNFetchBlob from 'rn-fetch-blob';
+
+class FileSystemStatic {
+    public getDocumentDir = () => RNFetchBlob.fs.dirs.DocumentDir;
```
vujevits: Maybe rn-fetch-blob has the functionality that we use in react-native-fs?

src/ImageDownloader.ts
```diff
@@ -0,0 +1,35 @@
+import sha1 from 'js-sha1';
+import { getSwarmGatewayUrl } from './Swarm';
+import { FileSystem } from './FileSystem';
+
+const getExtensionFromPath = (path: string): string => {
```
vujevits: I think we have quite a few places now which deal with path and filenames, maybe it's time to be consolidated in one place

src/RSSPostManager.ts
```diff
@@ -314,10 +315,11 @@ class _RSSPostManager {
             if (feedWithMetrics) {
                 downloadSize += feedWithMetrics.size;
                 const rssFeed = feedWithMetrics.feed;
-                const favicon = rssFeed.icon ? rssFeed.icon : await FaviconCache.getFavicon(rssFeed.url);
-                Debug.log('RSSPostManager: ', rssFeed, favicon);
+                const favicon = rssFeed.icon ? rssFeed.icon : await CachingFaviconParser.getFavicon(rssFeed.url);
```
vujevits: for higlighting the difference it may be better to name them localFavicon and remoteFavicon

android/app/src/main/AndroidManifest.xml
agazso: I installed the rn-fetch-blob package, maybe it added it.

src/FileSystem.ts
agazso: Yes, there is an overlap, rn-fetch-blob has filesystem functionality and react-native-fs also has download functionality :)

src/ImageDownloader.ts
agazso: I agree and I think the FileSystem could be a good place for that.

src/RSSPostManager.ts
agazso: There are two different things here:
- if the favicon is missing from the RSS feed, we try to get the favicon url by parsing the html page at the feed's url. In order to avoid parsing it every the user refreshes the feed we cache this information in the `CachingFaviconParser`.
- the point of this PR is once we have the url of the favicon, download them and store them in the filesystem, so that later we don't have to download it and it works offline as well. That's what the `ImageDownloader` does.

Also in my latest commits I dropped the idea of storing a separate localFavicon url, because it made the situation worse: everywhere we use the favicon we should have deal with a separate case of having a localFavicon or not etc.

Instead I chose to store the original favicon url (or authorImage in case of PostFeeds) as is, and now the Favicon component is using the ImageDownloader to check if the url is downloaded and if not download it and store it on the disk. 

From some perspectives it's less elegant because it totally avoids redux and the AppState and it makes a container responsible for modifying the stored state. On the other hand I think it's one of the limitations of redux, because it assumes every state data is stored in the AppState, and the lookup should be in a container. But in this case where the data is stored in the filesystem, you cannot use a container for the lookup because mapStateToProps cannot be async.

Also since this data we store is cached data (it can always be regenerated from the AppState) I don't think this is a huge issue.

agazso: Exactly that's what I plan to do. I had two problems with that earlier:
- It was difficult to add anything to the store (this will be fixed by #135 )
- We need to keep the app state caching consistent with the disk caching. Usually caching code is easy to get wrong (inconsistency) and now we are adding a second layer caching.

I tried to it in the component, but we had too many unhandled promises and that causes UX problems when it comes from the rendering code (stuck rendering, empty components etc.)

So yes, I will add the cache to AppState.

agazso:
This was included and merged in #135

vujevits:
this is still not merged to master, the better-redux-actions branch was branched from avatar-caching, merging that just merged it back here. 
https://github.com/agazso/postmodern/network

PR #132, opened 2019-01-19
remove immutable-js, add migration

### LINKED ISSUE #128: remove immutable.js
#### Problems
Causes several issues in our stack:
- oversized redux-persist store
- incorrect typing
- atm we are using it incorrectly (converting it to array at the wrong places, gaining almost nothing from immutability at the end of the day)

vujevits:
### [WIP / open for discussion]
- fixes #128 

### Issues
- can't get rid of immutable-js dependency, because it's needed for the migration (I guess we drop support after some time, private beta release being the latest)
- there is no way at the moment to only use a transform during migration, so the immutable-transform has to stay. see https://github.com/rt2zz/redux-persist/issues/686
- with all this being said, it actually works, migration included
- but if the immutable-transform is removed from the config, the rehydrated state becomes a string...
- a possible solution to this could be a 2 part migration, the caveat is that it has to be really 2 steps(application run with migration1, then update app and application run with migration2), not 2 consecutive migrations in 1 run (because of the changes to persistConfig)


src/helpers/immutable.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function replaceItemInArray<T>(arr: Array<T>, element: T, index: number): Array<T> {
+    return [...arr.slice(0, index), element, ...arr.slice(index + 1)];
+}
+
+export function updateArrayItem<T>(arr: Array<T>, index: number, cb: (item: T) => T): Array<T> {
+    return [...arr.slice(0, index), cb(arr[index]), ...arr.slice(index + 1)];
+}
+
+export function insertInArray<T>(arr: Array<T>, element: T, index: number): Array<T> {
```
agazso: This is generic insert function which is nice, but I think for lists we only need `push` like functionality, that is appending an element to the end.

vujevits: yes, it was written in another project originally just copied it over here.

agazso:
I created another branch from this branch, where I hacked out the necessary functionality to deserialize the immutable typed fields from the stored state.

With that we no longer need the `immutable` package, neither the `redux-persist-transform-immutable`.

https://github.com/agazso/postmodern/tree/remove-immutable-transform-hack

vujevits:
I merged that branch back here

PR #141, opened 2019-01-28
Fix refresh with reselect

vujevits:
The goal of this PR is to fix a bug when the refresh indicator disappears due to unrelated state changes. This is solved by using the reselect package to add memoization for filterings done in the view containers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "redux": "=4.0.1",
     "redux-persist": "=5.10.0",
     "redux-thunk": "=2.3.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -69,6 +69,7 @@
     "redux": "=4.0.1",
     "redux-persist": "=5.10.0",
     "redux-thunk": "=2.3.0",
+    "reselect": "^4.0.0",
```
agazso: We should use exact versions (`=` instead of `^`)

PR #142, opened 2019-01-29
Cli scripting

agazso:
Added CLI option to be able to access code from the app. Also added some scripts that demonstrates the usage.

Currently the CLI may only be used to print the version, however later we can add extra functionality to make backups, test different versions, server side scripting etc.

I tested it somewhat, but not for edge-cases. Tomorrow I have to release a new version, because the previous failed to upload to the AppStore and cannot upload the current version again.

### DIFF
--- a/scripts/cli.sh
+++ b/scripts/cli.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -e
+
+source="src/cli.ts"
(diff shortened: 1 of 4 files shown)


vujevits:
great!
I'm also happy we are already at version 0.9.x 😉

Issue #63, opened 2018-12-05
feeds are hard to find / access


vujevits:
### Ideas 
- Have a sidebar
- Search bar

david-gauquelin:
I think this is a very high priority issue, because it prevents people from using the app. If I have nothing to follow the whole app is meaningless and I won't engage with it. I see multiple opportunities at the moment.

**Easier to find/add people I know**
- ability to search by name, email or username
- invite people to my feed by email or message (from my dashboard I can type an email or enter a phone number and we send an invite email with info to add that person (ideally with link to the app but not necessarily)

**Easier to access feeds**
At the moment the first 3 tabs in the app are confusing for me. I understand the concept of separating different types of content and I'm not necessarily saying that it's not a good idea but I think there multiple opportunities to make it better:
- ability to mark feeds as favorites directly from Settings/Feeds
- ability to view feed from Settings/Feeds (not only the QR code)
- ability to view/manage a list of corresponding feeds directly from the News or Favorite page (view, remove, add, jump to...)

**Questions**
I'm not sure if the bottom tabs are the right place to separate feed, as an early idea I'm thinking that it might be a better idea to have a single feed tab and have a simple sub-menu/filter to isolate those feeds. Maybe we can even imagine combining more than just 1 category at a time. I would explore this idea

I'm not sure if it's a good idea to keep my posts in the first tab. I would rather expect to access this from my profile or something like this. It doesn't seem to be the most important thing and it's given the default, most prominent place atm.

agazso:
I agree this is very important. We have an onboarding label that we use for the issues that affects the users' first experience with the app. I added it to this issue.

I also created a new issue regarding making finding and adding people you know. Let's separate that discussion and in this issue focus on the feeds only.
https://github.com/agazso/postmodern/issues/139

We also already worked on the issues you mention (ability to mark feeds as favorites, ability to view feed from Settings/Feeds) and they are already merged and will be released in the next version (probably early this week).
https://github.com/agazso/postmodern/pull/127

So the current situtation with the tabs are definitely a result of an organic evolution and is not meant to be a final solution, just an experiment to try out the separation of different feeds until we build a filter component. So that seems like the logical next step to work on, because you independently reached that conclusion too.

[...]

PR #143, opened 2019-01-29
Find feeds

vujevits:
This is a minimal solution for issue #63 , but it might be useful at it's current state for the users. I would like to have something which resembles more of a dropdown or modal. Also it's only implemented for the news tab yet.
Might worth including it in the release.

### DIFF
--- a/src/containers/FeedListViewerContainer.ts
+++ b/src/containers/FeedListViewerContainer.ts
@@ -24,7 +24,7 @@ const mapStateToProps = (state: AppState, ownProps: { navigation: any }): StateP
 };
 
 const onPressFeed = (navigation: any, feed: Feed) => {
-    navigation.navigate('Feed', { feedUrl: feed.feedUrl });
+    navigation.navigate('Feed', { feedUrl: feed.feedUrl, name: feed.name });
 };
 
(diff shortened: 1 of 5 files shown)


agazso:
I like it, although I agree that a dropdown or modal would be better. But I think it's usable in it's current form and I would put it on the favorites tab too.

agazso:
I found one issue: if you go to the feed, favorite it then go back to the feed list, the recently favorited feed is not shown as favorite. You have to go back to the News feed and press the hamburger icon to bring in the updated feed list. Maybe it is connected to #141 ?

src/containers/FeedListViewerContainer.ts
```diff
@@ -0,0 +1,38 @@
+import { connect } from 'react-redux';
```
agazso: This file is almost a copy of FeedListEditorContainer. I expect lot of development happening around these files in the near future and this may result in introducing subtle but important differences how they behave. Please consider different methods of reusing more code.

agazso:
I merged the branch as is for trying out the current functionality in the release. However I did not delete the branch, so that we can continue discussing the issues that were raised.

PR #145, opened 2019-02-01
WIP Resharing posts including UI and network storage

agazso:


src/components/Card.tsx
```diff
@@ -91,9 +95,42 @@ const ActionIcon = (props: { name: string}) => {
     return <Icon name={props.name} size={iconSize} color={Colors.DARK_GRAY} />;
 };
 
+const isPostShareable = (post: Post, author: Author): boolean => {
```
vujevits: shouldn't this be a prop and calculated in the container?

agazso: Yes, that's a good idea

agazso:
> I was just thinking about this idea/problem, if our test privateKeys 'leak', then anyone can break our tests. Obviously only in unmocked (manual?) tests.

The tests should ideally work with a local isolated Swarm instance. We can add an option for generating test keys before running the tests in case of you want to do it on the public network.

PR #149, opened 2019-02-02
Add feed from clipboard

agazso:
When adding new feed, FeedInfo finds the content of the clipboard for a link and add automatically

### DIFF
--- a/src/components/FeedInfo.tsx
+++ b/src/components/FeedInfo.tsx
@@ -8,7 +8,6 @@ import {
     Dimensions,
     Clipboard,
 } from 'react-native';
-import Ionicons from 'react-native-vector-icons/Ionicons';
 import QRCode from 'react-native-qrcode-svg';
 import QRCodeScanner from 'react-native-qrcode-scanner';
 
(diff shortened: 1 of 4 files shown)


agazso:
I also changed the timeout of fetching a Swarm feed, because it often when adding a feed.

src/components/FeedInfo.tsx
```diff
@@ -192,9 +197,24 @@ export class FeedInfo extends React.Component<DispatchProps & StateProps, EditFe
...
+    private tryToAddFeedFromClipboard = () => {
+        const isExistingFeed = this.props.feed.feedUrl.length > 0;
+        if (!isExistingFeed) {
+            Clipboard.getString().then(value => {
+                const link = Utils.getLinkFromText(value);
+                if (link != null) {
+                    this.setState({
+                        url: link,
+                    });
+                    this.fetchFeed().then();
```
vujevits: we could use await and mark tryToAddFeedFromClipboard and componentDidMount async

agazso: 👍

agazso:
Changed the code to show the QR camera only after the clipboard was checked for links and no link were found. This avoids the confusion for first time users, who get a link from someone and then when trying to add the camera permissions dialog comes up without using the camera.

Also added some explaining text to describe what's going on (the wording could be improved here, especially if we find a better name for feeds).

vujevits:
At least with password managers I know it's a practice to flush the clipboard, I think we can argue that a feed url is sensitive data too.

agazso:
I added clearing the clipboard after the feed was successfully fetched. I think it works better this way and it's easy to disable if we change our minds later.

Issue #134, opened 2019-01-21
Make swarm gateway address configurable


agazso:
Then it can be changed easily to a public/private gateway or to the local lightnode address.

agazso:
This can be also useful for debugging #66, so it would be good to start with this issue.

agazso:
Also today the HTTPS certificate of the public gateways expired. 

We could operate two gateways, one for the swarm public testnet and an other one for a single private testnet.

agazso:
Timeboxed it to one day.

Issue #150, opened 2019-02-05
Coding conventions


agazso:
We should standardise coding conventions, the things that are not covered by TSLint.

We use different conventions in different files and it is very inconsistent sometimes. It can be very confusing when switching between different files adhering different standards. Here are some issues that I am aware of:

- const naming: we are trying to use const as much as possible, which is a good thing and makes the code more understandable. However in Javascript you end up using it for variable declaration, function declaration besides actual **constant** declarations. For variable names (including function names) we use `lowerCamelCase`, but for actual constant definitions we are using a mix of `UpperCamelCase` or `UPPER_SNAKE_CASE`.

- free functions vs. static members for utility functions: sometimes we put free functions in a file and directly export them (e.g. `random.ts`) and sometimes we create classes with static members and export the class (e.g. `Debug.ts`). I prefer the former (the latter is mostly very early code anyhow) but I like the 'namespacing' that comes with the latter, because it allows you to have shorter (and sometimes colliding) function names.

- file names: some of our file names start with uppercase, some of the starts with lowercase. There seems to be an organising force behind it: files that define types starts with uppercase, files containing only function definitions starts with lowercase, but there are some exceptions.

- classes vs. free functions & interfaces: most of the time we are using free functions, functional style. This style can be a bit cumbersome, for example when you want to wrap some context or configuration in several of your function calls. In these cases we use classes, but when using classes for pure functions you have to use `this` and that itself can be problematic in Javascript. I already experimented with a different style in `Swarm.ts` with the `FeedApi`: you implement an interface within a factory function, that can also extend the interface, store variables as closure etc.

vujevits:
I don't have a strong opinion anymore in a big portion of these questions, the reason is that many of these standards come from other languages, and in javascript the use-cases for them kind of mesh together / overlap.
- constant declarations: I think the `UPPER_SNAKE_CASE` is appropriate if it's used in several places, the usage is not tied to one file. If more then 2 files use them, ideally they can go to a separate file. Related topic: `readonly` is a TypeScript property modifier. It even disallows usage of mutating apis on arrays. Since `const` keyword is only useful for primitive types and forbidding reassignment, we could use it for more complex structures. There is also a type which makes its parameter's properties readonly. Eg.: 
```
type DefaultAppState = Readonly<AppState>; 
const defaultAppState: DefaultAppState = {....
```
Probably  not a good example but you get the idea.
- free functions: for namespacing we can use a convention at the import side (* as ...)
[...]

PR #151, opened 2019-02-06
Store and sync PostCommandLogs with Swarm

agazso:
Now we can store all post data in `PostCommandLog` locally and (in theory, hasn't yet well tested) then they can be synced automatically with other `PostCommandLog`s. The `PostCommandLog`s can be stored anywhere, for example on Swarm. This means it also becomes possible to sync between multiple devices.

There are helper functions to achieve this (e.g.
```ts
syncPostCommandLogWithSwarm: (postCommandLog: PostCommandLog, swarmApi: Swarm.Api): Promise<PostCommandLog>
```

Current issues:
- Right now I kept the implementation the most straightforward so it's not too efficient. For example right now there is no caching or logic to find quickly the last synced version, so we download the whole log and merge it with the whole log. The plan is to write tests for functionality first, then we can start thinking about optimizations.
- The `PostCommandLog` type now only consists a simple `PostCommand` array, so sometimes it feels extra work to pass around logs. On the other hand I can clearly see that in the future we can put extra data there, especially if we do conversions for other platforms and standards (e.g. ActivityStreams).

The next step are:
- [x] to make the Swarm API cleaner and more parametrisable (e.g. gatewayAddress).
- [x] create a separate Swarm Social layer which handles `PostCommandLog`s to keep Social API pure
- [x] cleanup and document Social API
- [x] write tests for edge-cases

After that is finished
- [x] when sharing a post use the Social API to post on Swarm
- [x] keep a `PostCommandLog` per feed on the device


src/swarm-social/swarmStorage.ts
```diff
@@ -0,0 +1,377 @@
...
+import { Utils } from '../Utils';
+import { PublicPost, Post, Author } from '../models/Post';
+import { ImageData } from '../models/ImageData';
+import { ModelHelper } from '../models/ModelHelper';
+import { MockModelHelper } from '../models/__mocks__/ModelHelper';
+import { Feed } from '../models/Feed';
+import { RecentPostFeed } from '../social/api';
+import { syncPostCommandLogWithStorage } from '../social/sync';
+
+export interface SwarmPostOptions {
```
vujevits: options is probably not a good name for this

src/swarm-social/swarmStorage.ts
```diff
@@ -0,0 +1,377 @@
...
+    imageResizer: (image: ImageData, path: string) => Promise<string>;
+    modelHelper: ModelHelper;
```
vujevits: I would only pass here the needed function(s)

agazso:
The next step from here:
- [ ] Add resharing functionality. This can be done on Post or PostCommand level. We may need to add extra properties to support that functionality.

Todos that needs to be fixed later:

- [ ] Syncing works only when assuming there is only one device doing the updates. In order to make it work with multiple devices we need to add to StorageSyncUpdate:
  - [ ] the list of posts that were not uploaded
  - [ ] the list of posts that were deleted
  - [ ] the device identifier as source
- [ ] If sync fails it throws an exception, but it's possible that a partial update was done. In order to handle failed uploads, we need the list above and to make sure the exceptions are caught during sync and handled properly.
- [ ] if the local PostCommandLog is updated (for example a new photo is shared during syncing) we try to do a sync as soon as possible and that can lead to a sync loop, especially when combined with the above issue. The solution is to implement a smarter retry mechanism.
- [ ] The current solution works only assuming a single feed, because topics are not set correctly.

src/actions/Actions.ts
```diff
@@ -177,94 +189,102 @@ export const AsyncActions = {
...
+                const ownFeeds = getState().ownFeeds;
+                if (ownFeeds.length === 0) {
+                    const identity = getState().author.identity!;
+                    const address = Swarm.makeFeedAddressFromPublicIdentity(identity);
+                    const author = getState().author;
+                    const ownFeed: LocalFeed = {
+                        posts: [],
+                        authorImage: author.image,
+                        name: author.name,
+                        url: `bzz-feed:/?user=${address.user}`,
```
vujevits: there should be a constant or helper function to assemble the url

src/actions/Actions.ts
```diff
@@ -177,94 +189,102 @@ export const AsyncActions = {
...
+                    if (originalPost != null) {
+                        const mergedImages = mergeImages(originalPost.images, updatedPost.images);
+                        dispatch(Actions.updatePostImages(updatedPost, mergedImages));
+                    }
+                    return updatedPost;
+                });
 
-                    const author = getState().author;
-                    dispatch(Actions.updatePostIsUploading(post, true));
+                // Re-check if there were an update to the command log during syncing
```
vujevits: will this still be an issue without gateway? (running swarm on mobile)

src/actions/Actions.ts
```diff
@@ -283,26 +303,13 @@ export const AsyncActions = {
...
-                    if (downloadedFeed.favicon !== '') {
-                        dispatch(InternalActions.updateFeedFavicon(feed, downloadedFeed.favicon));
-                    }
-                }
-            }
-        };
-    },
     restoreFromBackup: (backupText: string, secretHex: string): Thunk => {
-        return async (dispatch, getState) => {
+        return async (dispatch, getState: () => AppState) => {
```
vujevits: why was that needed?

src/actions/Actions.ts
```diff
@@ -332,15 +339,30 @@ const mergeImages = (localImages: ImageData[], uploadedImages: ImageData[]): Ima
...
         RSSPostManager.loadPosts(rssFeeds) as Promise<PublicPost[]>,
-        loadPosts(postFeeds),
+        loadRecentPosts(swarm, postFeeds),
     ]);
 
     const allPosts = allPostsCombined[0].concat(allPostsCombined[1]);
     return allPosts;
 };
+
+const getSwarmStorageSyncer = (signFeedDigest: Swarm.FeedDigestSigner, feed: LocalFeed) => {
```
vujevits: It might be better to have this in a separate file

src/reducers/index.ts
```diff
@@ -433,7 +441,7 @@ const appStateReducer = (state: AppState = defaultState, action: Actions): AppSt
 
 class FelfelePersistConfig implements PersistConfig {
     public transforms = [immutableTransformHack({
-        whitelist: ['contentFilters', 'feeds', 'ownFeeds', 'rssPosts', 'localPosts', 'postUploadQueue'],
+        whitelist: ['contentFilters', 'feeds', 'ownFeeds', 'rssPosts', 'localPosts', 'postUploadQueue', 'postCommandLog'],
```
vujevits: We shouldn't need the immutableTransformHack on postCommandLog

src/social/api.ts
```diff
@@ -1,124 +1,222 @@
...
+     * The actual post content
+     */
     post: Post;
-    source: string;
 
+    /**
+     * If the command is stored on Swarm, this is set to the {@link Swarm.Epoch}
+     * of the update
+     */
     epoch?: Swarm.Epoch;
```
vujevits: just a note for the future: 
to generalize this, we could pick some other name instead of epoch/previousEpoch, and using a union type when we want to support other platforms with different concepts

src/social/api.ts
```diff
@@ -1,124 +1,222 @@
...
+     * @returns the updates after syncing
+     */
+    sync: (postCommandLog: PostCommandLog, recentPostFeed: RecentPostFeed) => Promise<StorageSyncUpdate>;
+}
 
+/**
+ * Shares a new post and puts it in the command log
+ *
+ * @param post the post to be shared
+ * @param source the unique identifier of a user/device
```
vujevits: are there any privacy/security implications?

src/social/api.ts
```diff
@@ -251,237 +312,287 @@ export const removePost = (
...
-    await swarmFeedApi.updateWithFeedTemplate(feedTemplate, data);
-    return updatedCommand;
+/**
+ * Returns the epoch of the previous (first) command from the log
+ *
+ * @param postCommandLog the log
+ *
+ * @returns the epoch or undefined if the log is empty
+ */
+export const getPreviousCommandEpochFromLog = (postCommandLog: PostCommandLog): Swarm.Epoch | undefined => {
```
vujevits: [Don't use undefined as a means of denoting validity](https://basarat.gitbooks.io/typescript/docs/javascript/null-undefined.html)
I agree with this.

src/swarm/Swarm.ts
```diff
@@ -1,19 +1,19 @@
...
-import { safeFetch, safeFetchWithTimeout } from './Network';
-import { hexToByteArray, byteArrayToHex, stringToByteArray } from './conversion';
+import { PublicIdentity, PrivateIdentity } from '../models/Identity';
+import { Debug } from '../Debug';
+import { safeFetch, safeFetchWithTimeout } from '../Network';
+import { hexToByteArray, byteArrayToHex, stringToByteArray } from '../conversion';
 import { Buffer } from 'buffer';
 
-export const DefaultGateway = 'https://swarm.felfele.com';
+export const DefaultGateway = 'http://localhost:8500';
```
vujevits: is this for testing?

src/swarm/Swarm.ts
```diff
@@ -137,23 +137,44 @@ interface FeedTemplate {
...
-    Debug.log('downloadUserFeedTemplate: ', url);
-    const response = await safeFetch(url);
-    const feedUpdateResponse = await response.json() as FeedTemplate;
+const calculateFeedAddressQueryString = (address: FeedAddress): string => {
+    return `user=${address.user}` + (address.topic === '' ? '' : `&topic=${address.topic}`);
+};
+
+const downloadUserFeedTemplate = async (swarmGateway: string, address: FeedAddress): Promise<FeedTemplate> => {
+    const addressPart = calculateFeedAddressQueryString(address);
+    const response = await downloadFeed(swarmGateway, `bzz-feed:/?${addressPart}&meta=1`);
```
vujevits: as mentioned before, we interpolate / concat lot of swarm related urls, these should be handled at one place

src/swarm/Swarm.ts
```diff
@@ -166,70 +187,107 @@ export const updateUserFeed = async (feedTemplate: FeedTemplate, identity: Priva
...
         uploadPhoto: (localPath: string) => uploadPhoto(localPath, swarmGateway),
     };
 };
 
-export interface Api {
+export interface BaseApi {
+    readonly swarmGateway: string;
+}
+
+export interface WriteableApi extends BaseApi {
```
vujevits: Reader / Writer might be better.

vujevits:
👍 
well documentation too!

src/actions/Actions.ts
agazso: Yes, it's still an issue, just the probability is smaller. Because we wait after every feed update  one second, and there are minimum 2 feed updates when syncing, it still allows enough time to accidentally run into this and the result is a post that is stuck in the spinning state.

src/actions/Actions.ts
agazso: fixed

src/actions/Actions.ts
agazso: I was thinking the same, but then I couldn't decide where would I put it. I wouldn't put it under `swarm` or `swarm-social` because I think of them as separate generic packages and this code is specific to the app.

src/reducers/index.ts
agazso: fixed

src/social/api.ts
agazso: Yes, that was my idea too, just didn't wanted to generalise it now. Actually if we wouldn't use epochs just (bzz) links, it would be much flexible and less tied to Swarm. On the other hand we would lose the ability to sort by epoch, then we would have to provide a sorter function on the interface, so I just left it as is for now.

src/social/api.ts
agazso: Yes, with public posts it leaks the data what device you used to create the post. I was thinking of several ideas how this can be mitigated:
- If you know the devices ahead (because there is a pairing process for example) you can use some derivation from the device ids, that has some shared secret involved (e.g. hmac). In this case the device info is not publicly visible and if we add the timestamp to the value to be hashed it will change with every post. The downside is that to track the source we must calculate all the hash of the known device ids to identify the source.
- An alternative is to use the `privateKey` as a symmetric encryption key with the `timestamp` as nonce to encrypt the source. I read somewhere that this may not be a good idea, so further research is needed.
- Or actually we can just use random if we don't really care about which device created the data. The sorting/merging algorithm will then work correctly assuming there are no conflicting ids which may break the invariants, but if the random space is big then this is not an issue.

src/social/api.ts
agazso: I use this pattern at several places and at some places we use nulls as well, so let's discuss this further in the code conventions discussion #150 .

Also in the `PostCommand` when `epoch` is `undefined` it means that it hasn't been uploaded and this is consistent throughout the API. I am open to suggestions to improve this.

src/swarm/Swarm.ts
agazso: fixed

src/swarm/Swarm.ts
agazso: I agree, but I didn't want to spend too much time on the Swarm library, because I am already looking for a replacement (e.g. Erebos), because maintaining this is just too much effort.

src/swarm/Swarm.ts
agazso: I don't like Readable / Writeable either, right now my best idea is `DownloadApi`, `UploadApi`, `FeedDownloadApi` with `FeedUpdateApi`)

src/actions/Actions.ts
agazso: fixed

src/social/api.ts
vujevits: There are 2 things which come to my mind:
- it shouldn't be _the_ device id, or anything which is used for other then our application, because that would definitely mean a privacy leak
- maybe (after integration swarm light-node) we can use something specific to the node. That would satisfy the above requirement, and it's also different from device to device for the same user

src/swarm/Swarm.ts
vujevits: I'm good with Download(er) too.

agazso:
As far as I'm concerned we can merge it now to master. There are known issues:
- [ ] Resyncing can get stuck in a loop, now it waits 30 seconds before a resync to mitigate it somewhat
- [ ] I changed the deletion to update the PostCommandLog and sync, but this has not been tested yet

I added a screen to change the swarm gateway address, that closed #134.

vujevits:
I tried it, works, but there was a SyncPostCommandLog error quite a few times when I shared a post, BUT it still worked, managed to follow from another device. ATM cannot reproduce it, so let's merge.

PR #154, opened 2019-02-08
try to fix markdown hashtag bug

vujevits:
[WIP]
disclaimer: I'm not that good with regexp 🐱 
I would call this a naive solution. I added some tests, but there are 2 which are not working, my hunch is that it should be something obvious.


agazso:
This solution disables all the markdown escaping and implements an escape for only the hash, but it fails.

I suggest to try to write testcases for all the markdown special characters and try different versions of the original regex to understand how it works.

agazso:
Also the problem is most likely that we escape the markdown too early in the post editor. It should be only escaped when the user submits the post and not with every keystroke.

vujevits:
I'm not 100% sure what we are trying to solve here, the original issue was that the # changes the size of the text. The question is if we want to remove markdown editing altogether. I just realized that the hashtagging use-case does not really contradict the markdown functionality, because with markdown there has to be a space between the text and the # character.

agazso:
There are several issues here and the main idea is to remove the markdown editing altogether for now. We don’t give any indications that you can use markdown for editing and there is no preview either, so in order to avoid surprises when the users see their content rendered it’s the safest option. Later we may want to enable it for advanced editing options, but it also requires a smarter editor component.

The other thing is that we use a markdown library for react native to display the markdown text, which does a best effort rendering for different markdown formattings and it’s far from perfect. For example if you start a line with a hashmark it will automatically recognize it as text heading element, no matter if you put a space after or not. This is where the original issue was coming from.

For the hashtags I would also use a custom chooser component and not rely on raw user input.

vujevits:
👍

vujevits:
Now the referenced issue is solved, but many of the basic markdown functionality is still working:
- creating bulletpoint with dash and asterisk char
- italic text by surrounding with asterisk

I think his should be merged now because it solves the problem of the editor becoming unusable in case of inserting # char.

agazso:
Ok, but then create an other issue for the rest of the markdown issues 😊

agazso:
👍

PR #173, opened 2019-02-22
Post workflow

### LINKED ISSUE #79: Cannot take picture and post text from the app at the same time

### LINKED ISSUE #163: Post component on top of feed view

### LINKED ISSUE #164: create post flow

vujevits:
This PR fixes #164 and fixes #163 
Also an attempt which fixes #79 . It seems to be not working all the time (at least on android), it looks like the image has to be loaded in the post preview to later be displayed in the feed. Let's see it in practice if this is the case, I can only test on a very low-end phone.

### DIFF
--- a/src/ui/misc/TabBarPlaceholder.tsx
+++ b/src/ui/misc/TabBarPlaceholder.tsx
@@ -1,0 +1,12 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { DefaultTabBarHeight, Colors } from '../../styles';
+
+export const TabBarPlaceholder = (props: { color?: string }) => (
+    <View
+        style={{
+            height: DefaultTabBarHeight,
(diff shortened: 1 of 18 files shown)


src/ui/misc/TabBarPlaceholder.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { View } from 'react-native';
+import { DefaultTabBarHeight, Colors } from '../../styles';
+
+export const TabBarPlaceholder = (props: { color?: string }) => (
+    <View
+        style={{
+            height: DefaultTabBarHeight,
```
agazso: It seems this is not necessary on android, because there the TabBar is not transparent so the content is not behind it and this just puts a grey bar on the bottom of the screen.

Issue #91, opened 2018-12-21
Images should be saved in different resolution


agazso:
Optimally: 
- there should be a small image with almost only the background color, used as a placeholder (it can also be encoded as data for saving space and roundtrip time)
- then there should be another version for showing in the feed
- and there should be a high of full-resolution image uploaded

Download should start from the smallest size, then progressively download the second and on request download the third one.

PR #178, opened 2019-02-24
Image placeholder upload

agazso:
This implements part of #91, by uploading a 10x10 PNG image called `placeholder.png` to the manifest where we store images.

No changes on the download side yet.

### DIFF
--- a/src/swarm/Swarm.ts
+++ b/src/swarm/Swarm.ts
@@ -64,41 +64,42 @@ export const calculateTopic = (topic: string): string => {
     return '0x' + keccak256.hex(topic);
 };
 
-export const imageMimeTypeFromPath = (path: string): string => {
+type DefaultMimeType = 'application/octet-stream';
+
+export type MimeType =
+    | 'image/jpeg'
+    | 'image/png'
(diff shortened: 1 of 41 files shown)


src/swarm/Swarm.ts
```diff
@@ -78,19 +78,29 @@ const imageMimeTypeFromPath = (path: string): string => {
...
-    Debug.log('uploadPhoto: ', localPath);
+    return uploadPhotos([{localPath, name: 'photo'}], swarmGateway);
+};
+
+interface Photo {
+    localPath: string;
+    name: string;
+}
+
+const uploadPhotos = async (photos: Photo[], swarmGateway: string): Promise<string> => {
```
vujevits: I missed this earlier but calling these things Image/Picture is more descriptive, Photo is unnecessarily exclusive.

agazso: I renamed it to uploadFiles, because that's what it actually does.

https://github.com/agazso/postmodern/pull/178/commits/fdb1394ca980e83cbe31fcfc6f4ca153991bd767

PR #179, opened 2019-02-24
implement carousel component with react-swipeable-views

vujevits:
Very basic but useful swipeable image component, we could even make an option in the settings for autoplay.


agazso:
When the photos are not rectangular and there are both portrait and landscape photos in the collection there are lot of whitespace on the bottom and that doesn't look great.

I also assume that the rectangular photos is a legacy feature and we should remove the option for them.

vujevits:
suggestion?

agazso:
We can force to show the photos to be rectangular when there are more than one image. It's easy to implement now and later when we implement the separate image view, we can add smarter layouts for the timeline as a preview.

agazso:
Long tap is not working on the posts which have the carousel

vujevits:
@agazso I did not find any workaround for this, except for a `disabled` prop, when the swiping is disabled, and then the press/longPress works. Do you have any ideas with some UI tricks to make it user friendly? Basically we could turn on/off it somehow. Or disable it, and set it to autoplay.
Fyi if you longPress the dots, it works :-)

agazso:
Let's make autoplay then

PR #187, opened 2019-02-28
Splash screen

vujevits:
### Status
On my Android phone it does not look perfect.
On iOS Simulator it's a bit better. Let's merge it and see it in action.


agazso:
There is a visible white flash before loading on iOS. Let's investigate this further.

vujevits:
which one do you mean by before? 
- between clicking the icon and showing the splash screen
- showing the splash screen and showing the app

agazso:
Between the spash screen and the app. It’s possible that it’s because the Welcome component is empty, then we navigate to either the onboarding or the mainnavigator.

vujevits:
Good, we can try then to put the hide call to a later point, it just has to be handled on both branches (onboarding and normal)

vujevits:
I added the above mentioned code, but the real issue was hiding in the native part. 
There are 2 ways of adding the splash screen, without documentation on the difference, and the one I used first resulted in the white screen.

src/components/AllFeedScreen.tsx
```diff
@@ -24,29 +25,35 @@ export interface StateProps {
...
+                                    onSaveDraft={this.props.onSaveDraft}
+                                    profileImage={this.props.profileImage}
+                                    gatewayAddress={this.props.gatewayAddress}
+                                />,
+                }}
+            </RefreshableFeed>
+        );
+    }
+
+    public componentDidMount() {
```
agazso: Why did you choose to put it here? I understand that right now this is the first screen that we show if you are not onboarding, however this may change if we change navigation.

Did you try putting it into the App itself?

PR #189, opened 2019-02-28
Added autoplay to Carousel

agazso:


src/ui/misc/Carousel.tsx
```diff
@@ -28,7 +30,7 @@ export class Carousel extends React.PureComponent<Props, { index: number }> {
     public render() {
         return (
             <View>
-                <SwipeableViews onChangeIndex={this.handleChangeIndex}>
+                <AutoPlaySwipeableViews disabled onChangeIndex={this.handleChangeIndex}>
```
vujevits: Did not know this works with typescript now!

src/ui/misc/Carousel.tsx
```diff
@@ -57,6 +59,8 @@ export class Carousel extends React.PureComponent<Props, { index: number }> {
     }
 }
 
+export const AutoPlayCarousel = autoPlay(Carousel);
```
vujevits: This is not necessary, just use the carousel

PR #193, opened 2019-03-01
Changed default font everywhere

### LINKED ISSUE #190: Post text style
Change the style of the Markdown component in Card to match the one in the design.

agazso:
fixes #190


vujevits:
If we are going to use this, shouldn't the custom text components be removed?

PR #199, opened 2019-03-02
replace privateKey with OMITTED fixes #197

### LINKED ISSUE #197: Bugreport sends the privateKey in the mail

vujevits:
What about the publicKey / address? Should that be removed too?


agazso:
Ideally every personally identifiable information (PII) should be stripped of the bug report. That includes email address too, which is not possible now because we are asking a user to send a mail with the bugreport :)

So I would remove publicKey and address too, because linked with your email address it can be used to identify your feed. How about passing in an array of strings to the `getBugReportBody` with the data that needs to be omitted? This could include publicKey, address, but even your username or other personal data. For example your address is also included in the feed urls, so this could be used to omit that too.

PR #200, opened 2019-03-03
Felfele feeds

vujevits:
### WIP
- posted the default posts, it was working, right now both gateways say 'feed lookup fail: no feed updates found'
- nonetheless the posts should be available offline first as well
- started to experiment with creating a feed during onboarding, it seems to be working


agazso:
I would make the Felfele assistant posts local only, because it's part of the onboarding process and it should work when you are offline. Also I assume that the content we show depends on the users lifecycle which is different for everyone, therefore it makes little sense to have a unified online feed for the assistant. 

If we want to send updates and news, we can send it on the Felfele foundation feed.

agazso:
For the record, there is still a slight chance of feeds are not created:
- Because the app checks if the user is created by checking if there is an identity
- And because identity is created before ownFeed, because ownFeed needs values from identity
- It's possible that the identity is created, but the ownFeed is not yet (e.g. the app is restarted after the identity is created)

PR #206, opened 2019-03-04
Fixed opening own feeds, including the assistant, fixes #186

### LINKED ISSUE #186: Opening own feed does not work
Caused by this line:
https://github.com/agazso/postmodern/blob/master/src/containers/FeedContainer.ts#L25

state.ownFeeds is empty until something is shared (the feed is created)

related to #181

agazso:
I'm not completely happy with everything, however as far as I could see everything works, so even if we refactor some things we start from a working state.

Things we can improve:
- Change the AssistantFeed to a LocalFeed -> less special case handling
- NavigationHeaderButton: this already came up when the NavigationHeader was made, to create a component for a button that can has a `text` and `onPress` property. It would improve the readability of the NavigationHeaders.

### DIFF
--- a/src/containers/FeedContainer.ts
+++ b/src/containers/FeedContainer.ts
@@ -1,5 +1,5 @@
 import { connect } from 'react-redux';
-import { AppState } from '../reducers';
+import { AppState, defaultLocalPosts } from '../reducers';
 import { StateProps, DispatchProps, FeedView } from '../components/FeedView';
 import { AsyncActions, Actions } from '../actions/Actions';
 import { Feed } from '../models/Feed';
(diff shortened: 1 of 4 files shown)


src/containers/FeedContainer.ts
```diff
@@ -1,29 +1,44 @@
...
 import { getFeedPosts, getYourPosts } from '../selectors/selectors';
 
 export const mapStateToProps = (state: AppState, ownProps: { navigation: any }): StateProps => {
     const feedUrl = ownProps.navigation.state.params.feedUrl;
-    const selectedFeeds = state.feeds.filter(feed => feed != null && feed.feedUrl === feedUrl);
-    // TODO: we should have here state.ownFeeds.find(feed => feed.url === feedUrl) != null,
-    // but author.uri is empty, and the feedUrl from navparams is from post.author.uri (not real feedUrl)
-    // For the same reason feeds/selectedFeeds is empty (errounously)
-    const isOwnFeed = state.author.uri === feedUrl;
+    const feedName = ownProps.navigation.state.params.name;
```
vujevits: params.name could be more specific

src/containers/FeedContainer.ts
```diff
@@ -1,29 +1,44 @@
...
+
+    const isOwnFeed = feedName === state.author.name;
```
vujevits: we should store the author name in the feed

vujevits: or rather the feed should have an owner

src/containers/FeedContainer.ts
```diff
@@ -1,29 +1,44 @@
...
+    const isAssistantFeed = feedName === 'Felfele Assistant';
+    const hasOwnFeed = state.ownFeeds.length > 0;
+    const ownFeed = hasOwnFeed ? state.ownFeeds[0] : undefined;
+    const selectedFeeds = isOwnFeed
+        ? hasOwnFeed
```
vujevits: maybe this would be less ambiguous if we call this variable `isFeedSynchronized` or `isFeedSetUp` or...

src/containers/FeedContainer.ts
agazso: This is exactly how it is:
- we store the author of the feed in the `name` property of the feed (along with `url` and `feedUrl` properties)
- we have an `author` property in each post indicating the author of the post

src/containers/FeedContainer.ts
agazso: I didn't want to change this, because passing params through the navigation state is not type safe and I didn't want to introduce other breaking changes.

src/containers/FeedContainer.ts
agazso: I don't know what feed setup means in the context of the codebase. I know what feed synchronization means in the context of the codebase and it's not what we are trying to express. So to me the suggested names are more ambiguous.

vujevits: Yes these are not good either, I was just looking forint the correct wording. hasRemoteCounterpart is a bit long.

src/containers/FeedContainer.ts
vujevits: By design the idea is that the feedName != authorName

vujevits: That's true that we don't support that on the UI

Issue #191, opened 2019-03-01
Tests are very noisy


agazso:
I think ideally the tests should output nothing when there is no failure. Right now there are three sources of noise:
1. Console output
2. Duplicate mocks found by jest
3. Markdown errors in CardTest

The console output should be easy to fix as long as we agree that we only use logging for debugging. We can use this idiom in other tests for example:
https://github.com/agazso/postmodern/blob/master/test/social/syncTest.ts#L9

The other two requires further investigation.

agazso:
Fixed by https://github.com/agazso/postmodern/pull/210

Issue #209, opened 2019-03-05
Android build is broken with splash screen


agazso:
The error is the following:
```
/Users/attila/Projects/postmodern/android/app/src/main/res/layout/launch_screen.xml:5: AAPT: error: resource drawable/launch_screen (aka com.postmodern:drawable/launch_screen) not found.
    
error: failed linking file resources.

FAILURE: Build failed with an exception.
```

Full gradle error log:

```
% ./scripts/build_signed_android_release.sh
Loading dependency graph, done.
bundle: Writing bundle output to: android/app/src/main/assets/index.android.bundle
bundle: Done writing bundle output
bundle: Copying 8 asset files
bundle: Done copying assets
Configuration on demand is an incubating feature.

> Configure project :app 
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

> Configure project :react-native-get-real-path 
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

> Configure project :react-native-image-picker 
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html
WARNING: Configuration 'testCompile' is obsolete and has been replaced with 'testImplementation' and 'testApi'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html
WARNING: Configuration 'testApi' is obsolete and has been replaced with 'testImplementation'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

> Configure project :react-native-image-resizer 
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

> Configure project :react-native-securerandom 
WARNING: Configuration 'compile' is obsolete and has been replaced with 'implementation' and 'api'.
It will be removed at the end of 2018. For more information see: http://d.android.com/r/tools/update-dependency-configurations.html

> Task :app:bundleReleaseJsAndAssets 
warning: the transform cache was reset.
[...]

PR #210, opened 2019-03-05
run android release build on travis

vujevits:
- hopefully this will catch issues such as #209


agazso:
This PR also fixes #191

agazso:
Right now the android build is failing with the same error as locally.

The iOS build stops with this error:
```
error: There are no accounts registered with Xcode. Add your developer account to Xcode (in target 'postmodern')
error: No profiles for 'org.reactjs.native.example.postmodern' were found: Xcode couldn't find any iOS App Development provisioning profiles matching 'org.reactjs.native.example.postmodern'. (in target 'postmodern')
```

For example: https://travis-ci.org/agazso/postmodern/jobs/502523639

PR #225, opened 2019-03-14
Added device info to bugreport, fixes #223

### LINKED ISSUE #223: Add OS/version/device to Bugreport
**Describe the solution you'd like**
It would be good to add the device type, the OS and other software versions to the Bugreport.

agazso:


src/components/BugReportView.tsx
```diff
@@ -26,15 +29,37 @@ export const escapePII = (text: string, filterFields: string[]): string => {
         ;
 };
 
-const getBugReportBody = (filterFields: string[]): string => {
+const deviceInfo = () => {
+    const brand = DeviceInfo.getBrand();
+    const deviceID = DeviceInfo.getDeviceId();
```
vujevits: handling user/device ids is mentioned in the latest android beta blogpost, also references some best practices. It may be related:
https://android-developers.googleblog.com/2019/03/introducing-android-q-beta.html

agazso: Here the DeviceId refers to the model of a maker, e.g. iPhone 8 or Pixel XL3.

PR #230, opened 2019-03-15
News recommendation

### LINKED ISSUE #12: News recommendation onboarding
When you open the app for the first time, you can explore different predefined news sources based on your interest and the App will recommend you similar news feeds that you can follow.

vujevits:
### WIP

Fixes #12 

#### TODO
- [ ] json is local / static, upload it (LATER)
- [x] can't open a specific feed yet
- [x] serialization should be cached (reselect?)
- [ ] not part of onboarding yet (available from all feeds => explore) (#247 )
- [x] fix reddit loading, the issue is possibly with rate-limiting


vujevits:
I'm thinking about what would be an ideal workflow. Since what we are working with are not classical news-sites but tumblr / reddit, selecting based on an image and the title is probably not feasible, the users should open the individual feeds to decide about following. This is not implemented yet, just the grid with the images.

src/components/FeedInfo.tsx
```diff
@@ -243,9 +243,13 @@ export class FeedInfo extends React.Component<Props, FeedInfoState> {
             Debug.log('fetchFeedFromUrl', 'url', url);
             const canonicalUrl = Utils.getCanonicalUrl(url);
             Debug.log('fetchFeedFromUrl', 'canonicalUrl', canonicalUrl);
-            const feed = await this.fetchRSSFeedFromUrl(canonicalUrl);
-            Debug.log('fetchFeedFromUrl', 'feed', feed);
-            return feed;
+            try {
```
agazso: This `try` breaks the build and it's not necessary, because there is a try in `fetchRSSFeedFromUrl` already.

vujevits: yes, this was not planned to be in the PR, it was just for debugging

src/ui/screens/explore/SubCategoriesScreen.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
vujevits: Probably

agazso:
About the workflow: we definitely need to add a feed preview where the user can decide if they want to follow a feed.
Navigation should be straightforward, so going back from the preview should take you back to the feed list etc., no skipping levels.

src/models/recommendation/NewsSource.ts
```diff
@@ -0,0 +1,37 @@
...
+    name: string;
+    list: NewsSource[];
+}
+
+import data from '../../../news.json';
+
+type CategoryMap = { [name: string]: SubCategoryMap };
+type SubCategoryMap = { [name: string]: NewsSource[] };
+
+export const serializeData = (categoryMap: CategoryMap = data) => {
```
agazso: I think this is not a really good name, because one may associate it with serialization (https://en.wikipedia.org/wiki/Serialization) when in fact it transforms a `CategoryMap` to a list of Categories. Maybe `categoryMapToList` is better.

Also I think this function can be eliminated altogether if we already store the data in `news.json` in a list of categories, that is in the exact format how we want to use it later.

vujevits: right, it did something different previously, but still readJSON would have been better.

src/actions/Actions.ts
```diff
@@ -431,8 +431,9 @@ const mergeImages = (localImages: ImageData[], uploadedImages: ImageData[]): Ima
 };
 
 const loadPostsFromFeeds = async (swarm: Swarm.ReadableApi, feeds: Feed[]): Promise<Post[]> => {
-    const rssFeeds = feeds.filter(feed => !isPostFeedUrl(feed.url));
-    const postFeeds = feeds.filter(feed => isPostFeedUrl(feed.url));
+    const feedsWithoutOnboarding = feeds.filter(feed => feed.feedUrl !== 'local/onboarding');
```
agazso: Use `FELFELE_ASSISTANT_URL` instead of `local/onboarding`

src/reducers/defaultData.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export const defaultMetadata = {
+    highestSeenPostId: defaultLocalPosts.length - 1,
+};
+
+export const defaultFeeds: Feed[] = [
+    {
+        name: 'Felfele Assistant',
+        url: 'local/onboarding',
+        feedUrl: 'local/onboarding',
```
agazso: Use FELFELE_ASSISTANT_URL instead of 'local/onboarding'

agazso:
The padding is missing from the bottom of the category list, so I cannot access transportation.

agazso:
Some of the feeds are not working (for example the first one, Activities / Hobbies / Untitled). I would rename or throw away Untitled feeds. Some of the feeds are wrongly categorised (Health / Medicine / <pill> ) or almost empty (Health / Diets / Whoa).

With that said we can merge the functionality, but then we should create an other card for curating the content.

vujevits:
i already started to talk about it with @giurim

src/reducers/defaultData.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const defaultPost3: Post = {
+    _id: 2,
+    createdAt: Date.now(),
+    images: [],
+    text: `We have added some feeds that you follow automatically on the news tab (second tab). You can unfollow them if you don't like them. Enjoy!`,
+    author: onboardingAuthor,
+};
+
+export const defaultLocalPosts = [defaultPost1, defaultPost2, defaultPost3];
```
agazso: This should be called `defaultRssPosts` now

src/reducers/index.ts
```diff
@@ -428,7 +323,7 @@ class FelfelePersistConfig implements PersistConfig {
     public transforms = [immutableTransformHack({
         whitelist: ['contentFilters', 'feeds', 'ownFeeds', 'rssPosts', 'localPosts', 'postUploadQueue'],
     })];
-    public blacklist = ['currentTimestamp'];
+    public blacklist = ['currentTimestamp', 'exploreData'];
```
agazso: I think this is no longer needed

PR #232, opened 2019-03-16
Cli upload image command

agazso:
Fixes #200 by adding a favicon to the default Felfele feed. I added a gateway link because I spent lot of time with getting Node.js work (thus the CLI).

- Added `uploadImage` cli command
- It becomes possible to use the CLI for uploading and updating feed content


vujevits:
🚀 🍾 👍

PR #233, opened 2019-03-17
CLI without commander-js

agazso:
I removed `commander-js` although I liked the API originally. It had two major issues:
- It was not compositional, so if you wanted to create sub-commands with it, you had to write it with a switch/case, but then you lost `commander-js`'s support for help and other features.
- You could only check the values of the options after it started running the commands asynchronously. This led to race-conditions even with the very simple use-case we had, namely to turn Debug output off.

I was looking for alternatives and npm has at least 20 command line parsing projects, but most of them are very simple, one-off projects without tests and documentation, and `yargs` seemed too complex to use. I was thinking how hard can it be to write such a thing, so as a weekend project I built it :)
- It is fully compositional, so you can move commands between different levels very easily, with the options attached to it. 
- The interface is fluent and it lets you define options first, then commands but not the other way.
- It generates usage help that is very similar to commander, but it works also for subcommands.
- The whole library is less then 300 lines with no dependencies.
- It has almost the same amout of tests in LoC :)
- I plan to make it a separate npm package that we can use later and it will be good for practicing, because I've never done it before.

### DIFF
--- a/src/cliParser.ts
+++ b/src/cliParser.ts
@@ -185,7 +185,7 @@ interface OptionLookup {
 }
 
 interface Context {
-    name: string;
+    commandName: string;
     args: string[];
     definition: Definition;
 }
@@ -206,7 +206,7 @@ const readArgAndExecute = (context: Context, printer: Printer): Context => {
     if (isOption(args[0])) {
         const option = optionLookup[args[0]] || throwOptionError(args[0]);
         if (option === helpOption) {
-            printUsage(context.name, context.definition, printer);
+            printUsage(context.commandName, context.definition, printer);
             return {
                 ...context,
                 args: [],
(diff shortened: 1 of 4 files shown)


src/cliParser.ts
```diff
@@ -0,0 +1,283 @@
...
+const throwOptionMissingParameterError = (name: string) => throwError(`missing parameter to option \`${name}'`);
+const throwCommandError = (name: string) => throwError(`unknown command \`${name}'`);
+const throwCommandMissingParameterError = (name: string) => throwError(`missing parameter to command \`${name}'`);
+
+interface OptionLookup {
+    [name: string]: Option;
+}
+
+interface Context {
+    name: string;
```
vujevits: I assume this is command name

vujevits:
Look great! At some point it would be cool to move our CLI functionality to it's own package too :)

src/cliParser.ts
agazso: Yes, we can call it that.

PR #235, opened 2019-03-18
Feed update favicon

### LINKED ISSUE #208: feed update does not update favicon
**Describe the bug**
Creating a feed, and later changing the favicon does not update the favicon.

**To Reproduce**
1. Create feed
2. Add profile image
3. Share post
4. Feed image is not added / updated

**Expected behavior**
An update to the profile picture should happen either when adding the image, or when uploading a  new post.

agazso:
Supposed to fix #208, however as I started working on this I found lot of tech debt already:
https://github.com/felfele/felfele/issues/208#issuecomment-473972237

The original issue is that somehow the code that invokes the upload of the `Author` was removed, so we don't have a bzz link to the image and therefore everything is filled out with empty strings.

So far I started with removing `faviconUri` from `Author`, because this was legacy code, but we had to take it into account everywhere.

Maybe this will also touch topics described here, because there is already an overlap:
https://github.com/felfele/felfele/issues/181

### DIFF
--- a/docs/SocialApi.md
+++ b/docs/SocialApi.md
@@ -24,7 +24,6 @@ export interface PrivateIdentity extends PublicIdentity {
 export interface Author {
     name: string;
     uri: string;
-    faviconUri: string;
     image?: ImageData;
     identity?: PrivateIdentity;
 }
--- a/src/RSSPostManager.ts
+++ b/src/RSSPostManager.ts
@@ -438,7 +438,7 @@ class _RSSPostManager {
     private convertRSSFeedtoPosts(rssFeed: RSSFeed, feedName: string, favicon: string, feedUrl: string): Post[] {
         const links: Set<string> = new Set();
         const uniques: Set<string> = new Set();
-        const strippedFaviconUri = this.stripTrailing(favicon, '/');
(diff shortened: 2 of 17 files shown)


vujevits:
Shouldn't there be an action which uploads the profile picture if it changes? (also creating the feed if it did not happen previously)

agazso:
I considered that, but ultimately it quickly gets complicated if you think about what happens when the upload fails for example when there is no connectivity.

The implemented solution is that the profile picture is updated in the localFeed before the localFeed is  synced with Swarm (in `syncPostCommandLogs`). When the localFeed is uploaded (in `uploadRecentPostFeed`) we upload the profile picture if it were not uploaded before.

There is also the author added to each post and the profile picture is uploaded if it were not uploaded. Unfortunately this has the side effect that when you change your profile picture, post several posts and share them, but you are offline then when the posts are synced all of them will try to upload the profile picture. This means that it may need more data and time to upload the posts, but this is an edge-case and it still works correctly in the sense that the profile picture will be updated if the post is uploaded.

agazso:
The feed is automatically created when you finish the onboarding, but there is also this code when you share a post:
https://github.com/felfele/felfele/blob/feed-update-favicon/src/actions/Actions.ts#L279

PR #248, opened 2019-03-24
No any navigation

vujevits:
- make custom typing for the navigation object, to cover our needs
- typings work well, we can send params when navigating to, and read them on arrival in a typed manner
Caveats:
- it is verbose
- the usage of route names in App.tsx (where we have our config) are not covered by type checking (yet?)

### DIFF
--- a/src/helpers/navigation.ts
+++ b/src/helpers/navigation.ts
@@ -1,0 +1,69 @@
+import { ContentFilter } from '../models/ContentFilter';
+import { Feed } from '../models/Feed';
+import { SubCategory, NewsSource } from '../models/recommendation/NewsSource';
+import { LocalFeed } from '../social/api';
+
+export interface Routes {
+    App: {};
+    Loading: {};
+    Welcome: {};
+    Post: {};
+    Root: {};
+    ProfileTab: {};
+    PostTab: {};
+    FavoriteTab: {};
+    SettingsTab: {};
+    AllFeedTab: {};
+    BugReportView: {};
+    SwarmSettingsContainer: {};
+    FilterListEditorContainer: {};
+    Backup: {};
+    Restore: {};
+    BackupRestore: {};
+    LogViewer: {};
+    Debug: {};
+    FeedListViewerContainer: {
+        showExplore: boolean,
+        feeds?: Feed[],
+    };
+    Feed: {
+        feedUrl: string,
+        name: string,
+    };
+    FeedInfo: {
+        feed: Feed;
+    };
+    EditFilter: {
+        filter: ContentFilter,
+    };
+    FeedSettings: {
+        feed: LocalFeed,
+    };
+    FeedFromList: {
+        feedUrl: string,
+        name: string,
+    };
+    CategoriesContainer: {
+
+    };
+    SubCategoriesContainer: {
+        title: string,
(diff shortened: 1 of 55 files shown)


vujevits:
Possibly the typings could be improved to reduce the verbosity in the usage.

src/helpers/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+    NewsSourceGridContainer: {
+        newsSources: NewsSource[],
+        subCategoryName: string,
+    };
+    YourTab: {};
+}
+
+export interface TypedNavigation {
+    goBack: (routeKey?: string | null) => boolean;
+    navigate: <T, K extends keyof T>(routeKey: K, params: T[K]) => boolean;
```
agazso: This is quite verbose so I checked and everywhere we use this it is used with `T = Routes`. Therefore we could use this instead:
`navigate: <K extends keyof Routes>(routeKey: K, params: Routes[K]) => boolean;
`
Then using it would be as simple as:
`navigation.navigate('Loading', {});
`

src/helpers/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+        feed: Feed,
+    };
+    NewsSourceGridContainer: {
+        newsSources: NewsSource[],
+        subCategoryName: string,
+    };
+    YourTab: {};
+}
+
+export interface TypedNavigation {
```
agazso: I would call this simply `Navigation`

vujevits: I'd prefer to make the difference more apparent

src/helpers/navigation.ts
vujevits: 👍

vujevits: it would be nice to get rid of the {} as well, because that allows passing anything. (in case it is an empty object in the Routes)

agazso: Well, that's difficult to do in Typescript, because structurally every interface is equivalent with the empty interface.

src/helpers/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+    goBack: (routeKey?: string | null) => boolean;
```
agazso: Here the type of routeKey should be `K` and K should be a generic parameter `<K extends keyof Routes>`.

src/helpers/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+    navigate: <K extends keyof Routes>(routeKey: K, params: Routes[K]) => boolean;
+    pop: (n?: number, params?: { immediate?: boolean }) => boolean;
+    getParam: <R, P extends keyof R>(param: P) => R[P];
```
agazso: `R` generic parameter can be changed to `Routes`

src/helpers/navigation.ts
```diff
@@ -0,0 +1,69 @@
...
+    setParams: <R>(newParams: R) => boolean;
```
agazso: `R` generic parameter can be changed to `Routes`

src/helpers/navigation.ts
vujevits: I don't think we can do the same here. R here is not Routes, but a specific route, and we return a property of it.

Issue #54, opened 2018-11-29
Avatar is missing from shared posts


agazso:

vujevits:
is the edge case mentioned in #253 still an issue, or was it resolved lately by fixing other issues related to the avatar / images?

agazso:
Fixed by #330

PR #253, opened 2019-03-25
Added defaultImages, changed defaults everywhere

### LINKED ISSUE #182: There are lot of YellowBox warnings about `source.uri should not be an empty string`
I think it's related to the default profile picture, because it started after the redesign and if you don't set a profile picture it's empty in Settings.

agazso:
Fixes #182 .

- Added a new file containing the images that are bundled with the code with `require`
- Then they can be added as image sources by reference

Todos:
- [x] Now the image path is empty when you don't change the default image and it stays an empty string when the author is uploaded. It should be changed to either an already uploaded url or somehow the file should be uploaded.


src/ui/screens/feed-settings/FeedSettingsScreen.tsx
```diff
@@ -34,6 +35,10 @@ const UNLISTED_EXPLANATION = 'Anyone with a link to your feed can follow it.';
 
 export const FeedSettingsScreen = (props: Props) => {
     const modelHelper = new ReactNativeModelHelper(props.settings.swarmGatewayAddress);
+    const sourceImageUri = modelHelper.getImageUri(props.feed.authorImage);
+    const source = sourceImageUri !== ''
+        ? { uri: sourceImageUri }
+        : defaultImages.userCircle;
```
vujevits: offtopic, coding style 
I've seen that usually you put the semicolon at the end of a ternary ? : operator under the `:`, I prefer it at the beginning of a newline. I see that there is some aesthetics to it, but there are 2 cons in my opinion:
- it's the end of the operation, and usually we put that to the beginning of a newline (curly brace, closing parenthesis, etc.)
- it's also misleading because the `?` and `:` are nicely highlighted otherwise
Also I don't like write these things into stone, I've seen a very different way of writing a multi-level ternary conditional, which had a better readability in that case. 

This got a bit long and I'm not that emotional about it, just wanted to share my thoughts :-)

agazso: I was experimenting with different styles, we can stick to the semicolon in the beginning of a newline. Unfortunately there is no `tslint` rule for that.

agazso:
In order to not keep the `uri` empty in the uploaded posts when you don't choose a profile photo, I added the contents of the `images` folders as assets, so that they can be used from the app as images.

Unfortunately on android that's not enough, because the assets are stored in a compressed file (a zip file), so it also needs to be copied to the DocumentDir when we create create the profile. I don't really like this, but I could not find a better way to fix it.

This fixes #54, but I won't close that yet, because there is still an edge-case when uploading a post for the first time.

android/app/src/main/assets/custom/balloon.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vujevits: was there an ios issue being fixed here? Or why are the SVGs added here as well?

vujevits:
LGTM
It hurts a bit though...

agazso:
Since with this solution the files in the images were added twice to the app (once to the JS bundle, once to the native app) it increased the size of the app unnecessary.

I added a new folder called `images/assets` and only the files in it will be available from the native app. At the moment there is only the `user_circle.png` inside it.

PR #276, opened 2019-04-02
Fixed feed name and image change when changing profile name and image

### LINKED ISSUE #272: Profile and feed not matching
In the latest version (0.9.32) In the latest version I can have a different name/image for my profile and for my channel. Since we don’t allow to dissociate profile and channel for now, this is confusing.

agazso:
Fixes #272 .

- When updating the name or the image from the `IdentitySettings`(which should ideally be called `ProfileSettings`) it also updates the localFeed (if any).

### DIFF
--- a/src/containers/IdentitySettingsContainer.ts
+++ b/src/containers/IdentitySettingsContainer.ts
@@ -4,7 +4,6 @@ import { AsyncActions } from '../actions/Actions';
 import { StateProps, DispatchProps, IdentitySettings } from '../components/IdentitySettings';
 import { ImageData} from '../models/ImageData';
 import { TypedNavigation } from '../helpers/navigation';
-import { LocalFeed } from '../social/api';
 
 export const mapStateToProps = (state: AppState, ownProps: { navigation: TypedNavigation }): StateProps => {
     const ownFeed = state.ownFeeds.length > 0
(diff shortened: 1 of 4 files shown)


src/containers/IdentitySettingsContainer.ts
```diff
@@ -19,11 +20,24 @@ export const mapStateToProps = (state: AppState, ownProps: { navigation: TypedNa
 
 const mapDispatchToProps = (dispatch: any): DispatchProps => {
     return {
-        onUpdateAuthor: (text: string) => {
+        onUpdateAuthor: (text: string, ownFeed?: LocalFeed) => {
             dispatch(Actions.updateAuthorName(text));
+            if (ownFeed != null) {
```
vujevits: This looks fragile, it should be triggered by one action

src/containers/IdentitySettingsContainer.ts
```diff
@@ -19,11 +20,24 @@ export const mapStateToProps = (state: AppState, ownProps: { navigation: TypedNa
...
+                dispatch(Actions.updateOwnFeed({
+                    feedUrl: ownFeed.feedUrl,
+                    name: text,
+                }));
+            }
         },
-        onUpdatePicture: (image: ImageData) => {
+        onUpdatePicture: (image: ImageData, ownFeed?: LocalFeed) => {
             dispatch(Actions.updateAuthorImage(image));
+            if (ownFeed != null) {
```
vujevits: Same here

src/containers/IdentitySettingsContainer.ts
agazso: Changed

src/containers/IdentitySettingsContainer.ts
agazso: Changed

PR #290, opened 2019-04-04
Notifications

### LINKED ISSUE #251: Notifications
The simplest idea is to add a notification if you haven't used the app for a while.

Then the next idea is to add background downloading and fetch the felfele feed automatically and if there are updates create a notification.

agazso:
Fixes #251.

- Add local scheduled notifications to the app that can be triggered by some event or general inactivity
- Add background checking of a certain feed that can also trigger a local notification


android/app/build.gradle
```diff
@@ -135,6 +135,8 @@ android {
 }
 
 dependencies {
+    compile project(':react-native-push-notification')
+    compile project(':react-native-push-notifications')
```
vujevits: is this correct?

android/app/src/main/java/com/postmodern/MainApplication.java
```diff
@@ -3,6 +3,8 @@
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.dieam.reactnativepushnotification.ReactNativePushNotificationPackage;
+import com.dieam.reactnativepushnotification.ReactNativePushNotificationPackage;
```
vujevits: hm...

src/helpers/notifications.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Debug } from '../Debug';
+
+const localNotification = () => {};
+
+export const initializeNotifications = () => {
+    PushNotification.configure({
+
+        // (optional) Called when Token is generated (iOS and Android)
+        onRegister: (token: any) => {
+            Debug.log('TOKEN:', token );
```
vujevits: whitespace

src/helpers/notifications.ts
```diff
@@ -0,0 +1,60 @@
...
+        },
+
+        // (required) Called when a remote or local notification is opened or received
+        onNotification: (notification: any) => {
+            Debug.log( 'NOTIFICATION:', notification );
```
vujevits: whitespace

android/app/build.gradle
agazso: The readme says on android you have to link manually and I haven't worked on the android part at all yet. That's also the reason why the android build is broken.

agazso:
Resources:

RN lib for notification handling:
https://github.com/zo0r/react-native-push-notification

RN lib for background fetch
https://github.com/transistorsoft/react-native-background-fetch

RN lib for handling background activity (depends on the previous)
https://github.com/jamesisaac/react-native-background-task

Android background processing guide
https://developer.android.com/guide/background/

agazso:
So the current implementation has the following proposed changes:
- Added [react-native-push-notification](https://github.com/zo0r/react-native-push-notification) package for local and scheduled notifications. It works and it has an ok API, although the project seems dead. The alternative is [react-native-notifications](https://github.com/wix/react-native-notifications) but it's more (manual) work to set up, somehow you always have to check which platform you are using and I could not get it work in one day but it has recent commits so at least it's not dead.
- I wrapped the package API in `src/helpers/notification.ts`, we have to initialize the library by calling `initializeNotifications` and it will ask permissions from the user the first time it is called. It's also possible to separate the initialization from the permissions by changing the `requestPermissions` option.
- When the app is opened from a notification, an alert dialog is shown with the same text as the notification. We may want to change this in the future, but right now I implemented this way because ideally this should involve either 1) navigation, 2) badges or 3) both.
- Added [react-native-background-fetch](https://github.com/transistorsoft/react-native-background-fetch) for executing background tasks. Its name is misleading because it's not only for fetching in the background, but basically running arbitrary tasks, abstracted away by a common API. It's pretty well maintained, it even has `.d.ts` files and it uses reasonable implementations on both platforms.
- This package is wrapped in `src/helpers/backgroundTask.ts` with a quite straightforward function that sets reasonable defaults and takes care of releasing resources after the task is finished, so it doesn't have to be done manually every time.
- There is a new `AsyncAction` called `registerBackgroundTasks` that does two things:
  1. It sets up the background task by looking up the Felfele Foundation feed by name
  2. It also contains the implementation of the function that downloads the feed update, merges the new posts (if any) and sends a local notification about it.
This way by the time you open the app the updates will be already downloaded and possibly shown without extra user interaction.
- The background task is checked twice a day. I think it's a good balance between checking too often (thus draining the battery) and still getting updates in a timely manner, especially considering the way we want to send these notifications.
- There is a new `feed` subcommand of the `swarm` command in the CLI. It can be used to `create` feeds, `list` feeds and `post` new content. It requires to have an identity file which contains the `privateKey` to the feed. E.g.
```
./scripts/cli.sh swarm feed -i testIdentity.json post 'New post'
```
```
./scripts/cli.sh swarm feed -i testIdentity.json list
```
A test identity can be created with running the `./scripts/cli.sh swarm testId` and copying the json output to a file.
- Because adding these commands would have resulted in a too big file, I decided to split the `cli.ts` into separated files and move them into the `cli` folder.

agazso:
There is one known issue:
- the status bar icon on android is a grey box instead of the Felfele logo.

Screen Shot 2019-04-09 at 18 02 01

Screen Shot 2019-04-09 at 18 02 05

vujevits:
have you looked at https://github.com/zo0r/react-native-push-notification/issues/730 ?

PR #297, opened 2019-04-09
use the new server endpoint to post logs, made bugreportview stateful…

### LINKED ISSUE #274: bug report email only works with default email app on iOS
**Describe the bug**
If the Mail app is removed from the iPhone, it will only offer to download it from the App Store.

**To Reproduce**
1. navigate to report a bug
2. press send
3. popup asks you to download the mail app

**Expected behavior**
An email client should open

vujevits:
… to show an indicator

fixes #274 

- instead of an email, the bugreport is posted on our server
- a spinner is shown until the data is uploaded


vujevits:
server should work fine now, somehow in the committed code the content type was plain/text instead of text/plain

PR #300, opened 2019-04-10
Fixed default avatar bug

### LINKED ISSUE #264: Avatars are showing the default icon too often
**Describe the bug**
Normally the avatars that is shown for a feed is the icon of the feed, but too often it is the default icon (user in a circle in black and white).

This PR is responsible for this behavior:
https://github.com/felfele/felfele/pull/253

Because we changed that the defaults for the images but it was supposed to be a fix for the case when the image is not provided (e.g when keeping the default icon during onboarding), but instead it displays the default icon before (or instead) downloading the favicons.

agazso:
Fixes #264  .

- Improved the favicon parsing code and moved everything to FaviconCache
- Added heuristics to find the best available favicon
- The favicon is searched intensively when adding the feed
- Later when the feed is updated the stored value is used

### DIFF
--- a/src/FaviconCache.ts
+++ b/src/FaviconCache.ts
@@ -3,6 +3,11 @@ import { Debug } from './Debug';
 import { safeFetch } from './Network';
 import * as urlUtils from './helpers/urlUtils';
 
+interface Icon {
+    href: string;
+    size: number;
+}
+
 // tslint:disable:class-name
 class _FaviconCache {
     private favicons: Map<string, string> = new Map();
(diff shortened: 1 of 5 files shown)


src/FaviconCache.ts
```diff
@@ -60,21 +74,60 @@ class _FaviconCache {
...
+            return {
+                href,
+                size,
+            };
+        })
+        .filter<Icon>(isIcon);
+    }
+
+    private getBestIcon = (icons: Icon[]): Icon | undefined => {
+        const iconExtensionWeight = (iconHref: string) => iconHref.endsWith('.png')
```
vujevits: Can't we access the image sizes here somehow?
I'm not aware of the reason, but I think there are some cases when a png image uses slightly different extension pn + other character. It's only relevant because I think I've seen this in the case of tumblr or reddit.

agazso: Theoretically we could, if we downloaded the images and then opened them and queried the dimensions, but I think it's just not worth the effort. This case has been added because I found a website which had both `.ico` and `.png` without any other information and in that case the app should choose png anyhow.

It's possible for the website to provide different sizes of the icons (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) and we handle that case with the `getBestSizeFromAttribute` function.

If we found later that some websites use different image extensions than png, we can add them to this function easily later.

PR #301, opened 2019-04-10
implement new design for card post actions

### LINKED ISSUE #160: selected card post actions
- remove unimplemented actions
- how to display them (longtap is too advanced)

vujevits:
Fixes #160 

- overlay is shown instead of button row under the card
- only the two working buttons are present: share and delete

Issues
- feedback about uploading content / content being uploaded is not thought through yet

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -363,6 +363,7 @@ const styles = StyleSheet.create({
         flexDirection: 'row',
         height: 38,
         alignSelf: 'stretch',
+        alignItems: 'center',
         marginVertical: 14,
         marginLeft: 10,
     },
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -11,6 +11,7 @@ import {
     StyleSheet,
     Linking,
     Alert,
+    TouchableWithoutFeedback,
 } from 'react-native';
 import { TouchableView } from './TouchableView';
 import { DateUtils } from '../DateUtils';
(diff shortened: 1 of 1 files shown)


src/components/Card.tsx
```diff
@@ -50,29 +51,30 @@ type CardProps = StateProps & DispatchProps;
...
-                currentTimestamp={props.currentTimestamp}
-                author={props.author}
-                modelHelper={props.modelHelper}
-                navigation={props.navigation}
-                onSharePost={props.onSharePost}
-            />
-            <TouchableOpacity
-                activeOpacity={1}
-                onLongPress={() => props.togglePostSelection(props.post)}
-                onPress={() => openPost(props.post)}
```
agazso: Opening the post by tapping it was removed. I think this is a regression unless we add another option to open it (e.g. a new action button).

src/components/Card.tsx
```diff
@@ -261,6 +280,15 @@ const CardTop = (props: {
                 </View>
                 <RegularText style={styles.location}>{printableTime}{hostnameText}</RegularText>
             </View>
+            <TouchableView
+                style={{
+                    paddingRight: 10,
```
agazso: Some paddingTop is missing here.

Are it would be better to add this to the styles block instead of inlining according to our conventions.

agazso: The paddingTop is still missing :)

PR #306, opened 2019-04-11
Change add feed navigation

agazso:
I tested the app a lot manually and this always bothered me:

When you added a new feed
1. You press the plus button (Home screen)
2. Enter the name of the feed (FeedInfo screen)
3. The app opened the feed view (FeedView screen)
4. You go back to the (FeedInfo screen)
5. You go back to the Home screen

1. You press the plus button (Home screen)
2. Enter the name of the feed (FeedInfo screen)
3. The app opened the feed view (FeedView screen)
4. You go back to the Home screen

### DIFF
--- a/src/containers/FeedContainer.ts
+++ b/src/containers/FeedContainer.ts
@@ -28,7 +28,7 @@ export const mapStateToProps = (state: AppState, ownProps: { navigation: TypedNa
         : feedPosts
         ;
     return {
-        onBack: () => ownProps.navigation.goBack(null),
+        onBack: () => ownProps.navigation.popToTop(),
         navigation: ownProps.navigation,
         feedUrl,
         feedName,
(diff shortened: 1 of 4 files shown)


src/containers/FeedContainer.ts
```diff
@@ -28,7 +28,7 @@ export const mapStateToProps = (state: AppState, ownProps: { navigation: TypedNa
         : feedPosts
         ;
     return {
-        onBack: () => ownProps.navigation.goBack(null),
+        onBack: () => ownProps.navigation.popToTop(),
```
vujevits: replace could be another choice here. In both cases functionality is kind of fragile :-/

agazso: I don't necessary agree. Replace would be fragile, because it depends on the history and in our case we use FeedContainer on two tabs, so the two cases should be handled separately.

This just takes you to the toplevel, which is exactly what I wanted here. If we want something else we would have to touch this anyhow.

vujevits: I was not precise enough in my previous comment. Replace should be used when **navigating to** the FeedContainer. So the onBack could stay `navigation.goBack()`
This one is fragile in a sense that if you open a Feed elsewhere, the back button will take you somewhere 'unexpected'.

PR #321, opened 2019-04-18
added textinput so user can describe the bug in the report

### LINKED ISSUE #287: Bugreport redesign
Subtasks
- Change button style
- Fix send bug (#274)

https://app.zeplin.io/project/5c5332113af383308745c1b2/screen/5ca3748f78008b19f84837e0

vujevits:
Fixes #287 

- user can add feedback / bug description in the bugreport screen

Remaining questions:
- maybe we should leave the upper send button because it's easier to discover
- placing of the restart button is up to design, atm it's under the send button


agazso:
I found issues on iPhone 5s:
- There is a grey area on the bottom of the screen above the tab bar which has the same height as the tab bar and it is covering the content of the scrollable elements (marked with red in the screenshot):
Screen Shot 2019-04-18 at 17 51 32

- There is also the same grey area at the bottom of the PostEdit screen. I think it's because the `TabBarPlaceholder` is put at the end of the `FragmentSafeAreaView`, but that means that we have the placeholders everywhere, even when there is no tab bar. I see two possible fix ideas:
1. Pass the `TabBarPlaceholder` everywhere when it's needed as a children to the `FragmentSafeAreaView`. This is how I originally envisioned it to be used, but it may be problematic with screens that can be used with or without tab bar.
2. Add a `hasTabBar` property to the `FragmentSafeAreaView` and then depending on the value of the property add the placeholder or not.

agazso:
As a usability issue, generally it's not a good idea to put a scrollable component inside another scrollable component, as it is confusing which of the two are you scrolling etc.

I think in this case it would be better to put the log text at the bottom of the screen, below the Send button. Then the user could still review the content of the log, but I suspect most won't bother anyhow because it is technical. This way we wouldn't need another scrollable component.

@david-gauquelin wdyt?

vujevits:
discussed with @david-gauquelin and we both agreed with your proposition, implemented the necessary changes.

PR #334, opened 2019-04-29
News recommendation cleanup

### LINKED ISSUE #250: iterate on news-recommendation categories

agazso:
Fixes #250 .

- Renamed `news.json` to `exploreData.json`
- Cleaned up explore data, removed inactive feeds
- Removed internal conversion function and the needs for the `Category` and `SubCategory` types


vujevits:
why did you remove the loading indicator?

agazso:
Before this change `NewsSource`s were stored in the json, and in order to get a `Feed` to be displayed we had to load the feeds asynchronously and during that time it was necessary to show the load indicator.

Now we are storing `Feed`s in the json, there is no need for downloading and because of that there is no need to show the loading indicator either.

vujevits:
The images still have to be downloaded, no?

agazso:
Yes, but it's the same as with your channels. The list with the boxes can already be rendered, and then the image thumbnails will download shortly because they are usually quite small.

PR #335, opened 2019-04-29
Improve reddit feeds

agazso:
This depends on #334, that should be merged before this one.

- Added `redditFeedHelpers`
- When the app finds an RSS feed, instead it loads the json feed from Reddit
- It tries to find the best resolution image
- When there are no images there is also a link to comments

### DIFF
--- a/src/helpers/redditFeedHelpers.ts
+++ b/src/helpers/redditFeedHelpers.ts
@@ -50,9 +50,8 @@ const findBestResolutionRedditImage = (redditImage: RedditImage): RedditImageDat
 };
 
 const redditPostDataImages = (postData: RedditPostData): RSSThumbnail[] => {
-    const hasPreview = postData.preview != null;
-    const image = hasPreview
-        ? findBestResolutionRedditImage(postData.preview!.images[0])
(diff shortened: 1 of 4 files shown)


src/helpers/redditFeedHelpers.ts
```diff
@@ -0,0 +1,116 @@
...
+            height: [image.height],
+        }]
+        : []
+    ;
+};
+
+const redditPostDataToRSSItem = (postData: RedditPostData): RSSItem => {
+    const redditMobileLink = urlUtils.getCanonicalUrl('m.' + urlUtils.REDDIT_COM).slice(0, -1) + postData.permalink;
+    const created = Math.floor(postData.created_utc * 1000);
+    switch (postData.post_hint) {
```
vujevits: Why is this a switch and not an if?
Also the indentation and the usage of curly braces and returns is unusual.

agazso: Originally I thought the different values of the `post_hint` should be handled differently, but at the end only the undefined case is different.

I changed to an if.

PR #343, opened 2019-05-02
rename application to felfele in android and ios

### LINKED ISSUE #114: Change application name to felfele everywhere
Now in the app store it's called Last Offer. On android it's called postmodern and that's what shown on the splash page too. The .apk for android is called felfele.apk. This may be confusing.

Tasks to be done:

- [ ] Change the application name on iOS and android
- [ ] Create a new Felfele app in the Apple App Store
- [x] Change the splash screen with the logo and name

vujevits:
### DO NOT MERGE

Fixes #114 

- rename bundle name and application to Felfele


ios/Felfele.xcodeproj/project.pbxproj
```diff
@@ -1643,23 +1643,23 @@
...
+					"\"$(SRCROOT)/Felfele\"",
 				);
 				OTHER_LDFLAGS = (
 					"$(inherited)",
 					"-ObjC",
 					"-lc++",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)";
-				PRODUCT_NAME = postmodern;
+				PRODUCT_BUNDLE_IDENTIFIER = org.felfele.mobile;
```
agazso: I am not sure how it is supposed to work, but this used to include the `$PRODUCT_NAME` with an `rfc1034identifier` (whatever that means).

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "postmodern",
+  "name": "Felfele",
   "version": "0.1.0",
```
agazso: I think it at this point it would be a good idea to unify the versions as well. Currently on iOS we have 1.0.33, on android 0.9.33 and in the npm it will be present as 0.1.0.

vujevits: great idea :-)

agazso:
> Does it have to include the PRODUCT_NAME after all?

Well, it seems the bundle name has to include the product name (but not necessarily the `$(PRODUCT_NAME:identifer)` thing), because otherwise the App Store Connect will tell the bundle name is already registered, even though it is clearly not. This is a bug of the App Store Connect website.

agazso:
Fixes #346 too

PR #379, opened 2019-05-17
Universal links

### LINKED ISSUE #245: Deep linking
It would be great to have unified links for everything that could be used for deep-linking in the app.

For example we could have an invite link that would go to the website if the app is not downloaded or it would automatically open the feed in the app if it's installed.

Documentation

iOS: https://developer.apple.com/documentation/uikit/core_app/allowing_apps_and_websites_to_link_to_your_content
Android: https://developer.android.com/guide/topics/manifest/data-element.html

vujevits:
Fixes #245 

### Changes proposed in this pull request:
- app opens if someone taps on a link starting with *app.felfele.org*
- if the app is not installed, the user is redirected to the appropriate download site, implemented here: https://github.com/felfele/app.felfele.org

### TODO
- [x] open *Add channel* screen if the link contains a user address

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -36,6 +36,15 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <intent-filter android:label="Felfele">
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data
+                    android:scheme="https"
+                    android:host="app.felfele.org"
+                />
+            </intent-filter>
         </activity>
         <activity android:name="com.facebook.react.devsupport.DevSettingsActivity" />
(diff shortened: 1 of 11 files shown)


agazso:
What are the link formats? Can you provide some documentation?

vujevits:
the documentation is not very detailed for react-navigation (which we depend on in this case), but it should be something like this:
[baseUrl]/[screenPath]/[paramName]/[paramValue]
e.g.
https://app.felfele.org/channel/feedUrl/bzz-feed:/?user=0x1c3985e002bdc8e98731ede049354f63f0f6e2fe
If it's ready I can add proper documentation.

vujevits:
I made a few changes: the channels being shared are now base64 encoded, so it cannot mess up routing and parameters in the deep link. Also rss feeds and bzz feeds are handled the same way.
This is the link to follow index.hu rss feed:

Follow my channel on Felfele by opening this link:
https://app.felfele.org/follow/aHR0cHM6Ly9pbmRleC5odS8yNG9yYS9yc3Mv

The wording at this point is a bit misleading, because it's not my channel, not sure though how or if we should solve this. Otherwise I

PR #396, opened 2019-05-29
Feed view refactor

vujevits:
Get rid of feeds array


agazso:
Looks good, but since I merged the FeedView refactor it's in conflict now.

PR #400, opened 2019-06-04
270 change post editor

### LINKED ISSUE #270: Change post editor
Subtasks:
- Images on top, text below
- Images have delete button
- Placeholder for photo
- Allow maximum 3 photos
- Drag'n'drop photo ordering
- Show selected photos (#110)

vujevits:
Fixes #270 

- Images on top, text below
- Images have delete button
- Drag'n'drop photo ordering

We decided to drop the 3 photos requirement, the interaction is good enough for handling scroll. Placeholder would probably require too much work, the built in solution only works for vertical lists.


agazso:
Some issues that I see when trying it out
- Drag'n'drop is not working
- When there are <= 3 items, it should not scroll (maybe that's why the drag'n'drop is not working?)
- When the text has focus then you have to tap twice to delete an item

agazso:
Ok, I figured out how the drag'n'drop works (you have to longtap first, then you can move), but (at least on iOS) it feels very different from how the platform works normally. Usually that's the case when a component is implemented in pure javascript without using platform-specific components.

I would consider using a different component or not using it at all.

vujevits:
I could not reproduce tap twice to delete, although the hitslop area might be too big in this case. Is there a best practice for that? 

About the drag'n'drop on iOS, how does it work "normally"? The only interaction where I use it it is moving icons on the homescreen, - which is a pretty gimmicky implementation in my opinion -, but long pressing is the way to enter it.

agazso:
Regarding the hitslop, usually I test it on my phone until it feels "right" :) I don't know best practices but in my experience the hitslop size should be proportional to the size of the button and there should be a minimal size (e.g. 20x20) that it's still possible to tap. Usually problems arise when the hitslops are overlapping.

For reproducing the tap twice to delete:
- Enter some text in the editor so that the texteditor has focus
- Tapping first on the delete button just loses the focus of the text editor
- Tapping again the delete button deletes the image

This is related to this issue: https://github.com/felfele/felfele/issues/183
I think ideally the keyboard should be never closed, the focus should always stay in the text editor but you should be able to perform actions on the images.

As for the drag'n'drop, I tried it on the device and you are right right that it works similar to how it behaves on the homescreen. I agree that moving the icons on the iOS homescreen is gimmicky, maybe that's why I don't like this too much either :/

One more issue I found when testing on the device: after you reorder the elements there is a white flashing for the whole ImagePreviewGrid (except when you move to the last position).

vujevits:
Yes unfortunately there is a blinking issue when adding as well, there is an open issue about it here https://github.com/gitim/react-native-sortable-list/issues/143

I'll make the hitslop smaller.

About the drag'n'drop, we have kind of a big freedom regarding the animation, if we have time we can iterate on it later.

agazso:
Is there something left or can we merge this?

PR #405, opened 2019-06-06
Fix photo permission rejection on iOS

### LINKED ISSUE #389: If you reject permission to photos once, it will never be asked again
**Describe the bug**
If you reject permission to photos once, it will never be asked again. This means that any access to the photo library or the camera will do nothing.

**To Reproduce**
1. Reset the app
2. Tap the profile chooser in the onboarding
3. When the app requests permission choose 'Don't permit'
4. Trying to choose another profile image will do nothing

**Expected behavior**
If the permission was rejected, then the permission dialog should be showed again when the user taps anything that requires access.

### LINKED ISSUE #404: When changing the profile picture it's not possible to choose the camera
**Describe the bug**
When changing the profile picture it's not possible to choose the camera, only a photo from the photo library.

**Expected behavior**
When changing the profile picture it should be possible to choose the camera besides the photo library.

agazso:
Fixes #389  .

- Added an alert when permission is rejected, explaining how the settings can be changed
- Also fixes #404

### DIFF
--- a/src/AsyncImagePicker.ts
+++ b/src/AsyncImagePicker.ts
@@ -1,5 +1,5 @@
 import { ImageData } from './models/ImageData';
-import { Platform } from 'react-native';
+import { Platform, Alert } from 'react-native';
 import { Debug } from './Debug';
 import ImagePicker from 'react-native-image-picker';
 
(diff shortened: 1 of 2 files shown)


agazso:
Can you describe the steps to reproduce? You may need to 'Erase all Content and Settings' in the Simulator before you try it out.

The only change the new code does is that when the permission is rejected it shows an alert on iOS. That should not affect showing the permission dialog, which is done by the OS when the app tries to access it the first time.

agazso:
Ok, I think I figured out what happens. In the Simulator you are not allowed to use the camera at all, only the image library. For the app it seems that there is no permission to use the camera, so it shows the dialog. This is the correct behavior.

PR #416, opened 2019-06-16
Added invite flow test to CLI

agazso:
Instead of describing the invite flow in a PR, first I wanted to experiment with a working version to see the different tradeoffs and potential issues.

The code itself is working and servers as a Proof-of-concept version of the invite process with QR codes and links. I found it useful to model user flows, however in order be useful for describing different kind of flows in the future this code should be generalized, because right now it's too specific about the invite flow.


agazso:
Related #417

PR #418, opened 2019-06-17
Invite contacts with QR code

### LINKED ISSUE #417: Making contacts with QR code
**Is your feature request related to a problem? Please describe.**
Currently in the app we can only follow channels and that creates a one-way connection and that enables only pull based information flow. Therefore it's not trivial to send or share information selectively with a specific person(s).

**Describe the solution you'd like**
By introducing the concept of Contacts, we could setup two-way connections. For example exchanging public keys would make sure that both parties follow each other and they could setup a shared, encrypted channel to share information.

After there is established trust, later this schema can be extended to any number of participants in a private group.

agazso:
Fixes #417.

- Added Contact model type
- Have a list of contacts
-

### DIFF
--- a/src/helpers/contactHelpers.ts
+++ b/src/helpers/contactHelpers.ts
@@ -5,19 +5,25 @@ import { HexString } from './opaqueTypes';
 import { byteArrayToHex } from './conversion';
 import { ec } from 'elliptic';
 import { keccak256 } from 'js-sha3';
+import { Utils } from '../Utils';
 
 const stripHexPrefix = (hex: string) => hex.startsWith('0x')
     ? hex.slice(2)
     : hex
 ;
 
-const publicKeyToAddress = (contactPublicKey: HexString): HexString => {
+const publicKeyToAddress = (publicKey: HexString): HexString => {
     const curve = new ec('secp256k1');
-    const contactPublicKeyWithoutPrefix = stripHexPrefix(contactPublicKey);
-    const publicKeyPair = curve.keyFromPublic(contactPublicKeyWithoutPrefix, 'hex');
+    const publicKeyWithoutPrefix = stripHexPrefix(publicKey);
+    const publicKeyPair = curve.keyFromPublic(publicKeyWithoutPrefix, 'hex');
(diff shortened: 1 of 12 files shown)


src/helpers/contactHelpers.ts
```diff
@@ -0,0 +1,199 @@
+import { Contact, InvitedContact, CodeReceivedContact, AcceptedContact, MutualContact } from '../models/Contact';
```
vujevits: This kind of helper functionality is very feature specific,  I think this should go under `src/contact`. This is also true for some already existing code in the `helpers` dir.

agazso: You want to create a new folder to contain one file? Can you elaborate a bit on this?

agazso:
> How should we handle when someone deletes a contact? Do we delete it on the other-side as well?

When someone deletes a contact they no longer want to communicate with the contact. Deleting it on the other side would require communicating the fact that the contact was deleted _after_ the request of stopping communicating with the contact. So I wouldn't delete it on the other side.

vujevits:
This would give the illusion for the deleted user that she can still contact her peer in their private feed (of 2 of them) / messages. The bidirectional relationship suggests a "friends" status, I think that is not good UX of listing someone as such, if it's not accurate. I can imagine though more subtle differences, if we want to distance ourselves from this "friends" status.
- Keep it called everywhere a contact
- The action of sending something to the contact can be removed (reflected on the UI). Imagine a contact card, where there is no option of sending anything, just displaying name / image, and opening a public feed if there is such feature. If there is no other meaningful feature to display (e.g. a shared phone number, or email address), I find it difficult to argue why we should not delete it from the contact list

agazso:
I agree that it should not be called friend (because it might give the impression that it works similarly like FB) but rather contact or something else.

I think we can come up with different ways to implement this. What I suggested is the simplest to implement and also that preserves privacy the best. It is also similar in a way to your contact list on your phone in a sense, that you won't get notified when someone deletes your phone number and puts you on a blacklist.

vujevits:
You just articulated even better my concern:
I don't want go give the impression that you are deleting a contact, when in reality it's banning

PR #423, opened 2019-06-27
add action to update feed data

### LINKED ISSUE #313: Name/profile picture update is not updated in your feed list
**Describe the bug**
Name/profile picture update is not updated in your feed list after a refresh.

**To Reproduce**
1. Use two devices
2. `One` follows `Two`, so `One` has `Two` in the feed list
3. `Two` changes its name/profile picture
4. `One` sees the old name/profile picture after a refresh

**Expected behavior**
The updated name/profile picture should be shown after a refresh.

vujevits:
### Proposed changes
Fixes #313 
- when downloading posts, the feeds are updated as well

### Note
The code is not really efficient, probably we should move to using map / object for storing feeds.

### DIFF
--- a/src/actions/Actions.ts
+++ b/src/actions/Actions.ts
@@ -180,7 +180,9 @@ export const AsyncActions = {
             // TODO this is a hack, because we don't need a feed address
             const swarm = Swarm.makeReadableApi({user: '', topic: ''}, getState().settings.swarmGatewayAddress);
             const allPosts = await Promise.all([
-                loadSocialPostsFromFeeds(swarm, feedsWithoutOnboarding, dispatch),
+                loadSocialPostsAndUpdateFeeds(swarm, feedsWithoutOnboarding, (recentPostFeeds: RecentPostFeed[]) => {
+                    dispatch(InternalActions.updateFeedsData(recentPostFeeds));
+                }),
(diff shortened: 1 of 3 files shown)


src/actions/Actions.ts
```diff
@@ -172,10 +176,14 @@ export const AsyncActions = {
     downloadPostsFromFeeds: (feeds: Feed[]): Thunk => {
         return async (dispatch, getState) => {
             const previousPosts = getState().rssPosts;
+            const feedsWithoutOnboarding = feeds.filter(feed => feed.feedUrl !== FELFELE_ASSISTANT_URL);
             // TODO this is a hack, because we don't need a feed address
             const swarm = Swarm.makeReadableApi({user: '', topic: ''}, getState().settings.swarmGatewayAddress);
-            const downloadedPosts = await loadPostsFromFeeds(swarm, feeds);
-            const posts = mergeUpdatedPosts(downloadedPosts, previousPosts);
+            const allPosts = await Promise.all([
+                loadSocialPostsFromFeeds(swarm, feedsWithoutOnboarding, dispatch),
```
agazso: I would call this function something else. Also I wouldn't pass `dispatch` here, because we use it as `any` but a function that updates the feeds.

PR #424, opened 2019-06-27
Contact invite with UI

agazso:
The goal is to be able to add contacts with QR code or invite link and they should behave the same way they behaved with public feeds. The only visible difference is that the connection is created on both sides.

- [x] When adding a new contact, we add them to the channel list and display their public posts there
- [x] When the associated channel is deleted then the contact is deleted
- [ ] It can be added to favorites too (LATER)

Issues:
- [x] The profile name and images is missing from the invite protocol, therefore it's missing from the contact information too
- [x] Follow link is broken, because there was a fix to make testing easier but it broke deep linking #422 
- [x] Invite link deep linking is not yet implemented
- [x] Zero topic potentially breaks existing bzz-feed urls
- [x] The feed of the contact is twice in the channel list (once as contact, once as followed feed)
- [ ] Sometimes when displaying the QR code the polling is broken; the tab has to be changed and then changed back again to make it work (LATER)
- [ ] Non-interactive use-case with invite link or failed QR code scanning (LATER)
- [x] Invite link shown as two links in iMessage (Simulator issue)
- [x] The same contact can be added twice


vujevits:
`Invite link shown as two links in iMessage`
I think this is a simulator issue

agazso:
Fixes #417

src/components/FeedInfo.tsx
```diff
@@ -210,7 +189,72 @@ export class FeedInfo extends React.Component<Props, FeedInfoState> {
...
+                feedUrl: feed.feedUrl,
+                name: feed.name,
+            });
+        } else {
+            this.onFailedFeedLoad();
+        }
+    }
+
+    private async handleInviteCode(inviteCode: InviteCode) {
+        const swarmContactHelper = createSwarmContactHelper(
```
vujevits: `swarmContactHelper` could be a prop of the component as well

src/selectors/selectors.ts
```diff
@@ -31,6 +35,18 @@ const getSelectedFeedPosts = (state: AppState, feedUrl: string) => {
         });
 };
 
+export const getContactFeeds = createSelector([ getContacts ], (contacts) => {
+    const mutualContacts = contacts.filter(contact => contact.type === 'mutual-contact') as MutualContact[];
```
vujevits: I've found a way to use type guards for arrays too:
```TypeScript
const mutualContacts = contacts.filter((contact): contact is MutualContact => contact.type === 'mutual-contact')
```
I believe this should work here
See https://github.com/microsoft/TypeScript/issues/7657

PR #425, opened 2019-07-01
React native 0.59

vujevits:
### Changes
Fixes broken android build

- upgrade react-native (also prerequisite to #388)
- upgrade gradle
- qr code svg
Some other dependencies related to compilation and building

### Issues
An issue with style typing is back, similar to https://github.com/DefinitelyTyped/DefinitelyTyped/issues/29265


trickeyd:
I'm on .59.10 now and getting the error you mentioned, although it seems more sporadic than it was when it was happening with 57. Did you ever fine a combo of typscript / @types/react-native that fixed the problem?

vujevits:
@trickeyd no, still using a workaround for it. Haven't investigated it further, possibly it's an issue with the dependency `react-native-qrcode-scanner`, the only place I have this problem is with `QRCodeScanner`'s `containerStyle`.

trickeyd:
@vujevits Okay thanks! 👍  
I'll work around too for now then. I imagine it will be sorted when I finally make it to 0.60+

PR #432, opened 2019-07-04
rework clipboard handling in FeedInfo

### LINKED ISSUE #395: Rework auto clipboard pasting from add channel
False positive case:
- user has something on the clipboard, it starts with an error
- we delete the clipboard content without knowing what it was

Deep linking makes it a redundant feature

vujevits:
Fixes #395 

- fixed an issue when typing a url without protocol (e.g. index.hu) resulted in a network error
- automatic clipboard pasting & flushing is only done with felfele / swarm related links
- added some unit tests


src/components/FeedInfo.tsx
```diff
@@ -212,8 +213,15 @@ export class FeedInfo extends React.Component<Props, FeedInfoState> {
             this.handleInviteCode(inviteCode);
             return;
         }
-        const feedUrl = link;
-        this.handleFeedUrl(feedUrl);
+        if (urlUtils.isFelfeleResource(link)) {
+            this.handleFeedUrl(link);
+            return;
+        }
+        const feedUrl = urlUtils.getCanonicalUrl(link);
```
agazso: While this fixes the bug with the links without protocol, this is maybe not the best way to solve it:
1. The `if` is not really necessary, because `getCanonicalUrl` should never return `undefined` or `null`
2.  The root cause of this problem is in the `fetchFeedFromUrl` function from `feedHelpers.ts`, where we also get the canonical url, but then don't use it.

https://github.com/felfele/felfele/blob/5d8aed38849d293a76afa5703e0c2f5dfdac43f7/src/helpers/feedHelpers.ts#L50

I would fix there, because that function we use in other code as well.

src/helpers/urlUtils.ts
```diff
@@ -93,29 +98,33 @@ export const getLinkFromText = (text: string): string | undefined => {
...
 export const getBzzLinkFromText = (text: string): string | undefined => {
-    const bzzLink = text.match(/(bzz:\/\/[a-f0-9]{64})( |$)/);
+    const bzzLink = text.match(BZZ_RESOURCE_MATCHER);
     if (bzzLink != null) {
         return bzzLink[1];
     }
     return undefined;
 };
 
+export const isFelfeleResource = (text: string): boolean => {
```
agazso: We don't use `BZZ` resources as links, only `BZZ_FEED` resources.

src/components/FeedInfo.tsx
vujevits: fixed in 8442e7e

src/helpers/urlUtils.ts
vujevits: fixed in 15b9c82

PR #433, opened 2019-07-05
sharing to felfele from other apps

### LINKED ISSUE #369: Share to Felfele from other apps

vujevits:
### Description
Fixes #369 
Users should be able to share content through the native share popup which can be opened from anywhere on iOS and Android. 

### Tasks
- [x] add `react-native-share-extension` library
- [x] implement a screen for sharing content
- [x] solve runtime dependency on backgroundTask

#### iOS
- [x] add capability for basic functionality for link, text, image sharing
- [x] create App Group to access AppState
- [x] rework how we use redux-persist / AsyncStorage
- [x] create migration for new storage

#### Android
- [x] add capability for basic functionality for link, text, image sharing
- [x] debug crash on android


src/share-extension/ShareExtension.tsx
```diff
@@ -27,13 +29,22 @@ export default class FelfeleShareExtension extends React.Component<{}, ShareStat
...
         }
+        Debug.log('will render');
         return (
             <TopLevelErrorBoundary>
                 <Provider store={this.state.store!}>
                     <PersistGate loading={null} persistor={this.state.persistor!}>
-                        <SharePostEditorContainer text={this.state.value}/>
+                        <SharePostEditorContainer
+                            text={this.state.value}
+                            goBack={() => {
```
agazso: If I understand correctly, using `goBack` also means if you cancel posting anything from the ShareExtension and that results also in an update in the store.

Unfortunately the app start time is quite long so it would be good to minimize when it happens.

src/App.tsx
```diff
@@ -385,10 +388,17 @@ const InitialNavigator = createSwitchNavigator({
...
 
-export default class App extends React.Component<{}, { store: any, persistor: Persistor | null }> {
-    public state = {
+interface ApplicationState {
+    store: any;
+    persistor: Persistor | null;
+    nativeAppState: AppStateStatus;
+}
+
+export default class App extends React.Component<{}, ApplicationState> {
```
agazso: We could call this `FelfeleApp` and its state to just `State` or `FelfeleAppState`.

src/App.tsx
```diff
@@ -407,10 +417,33 @@ export default class App extends React.Component<{}, { store: any, persistor: Pe
     }
 
     public async componentDidMount() {
+        AppState.addEventListener('change', this.handleAppStateChange);
```
agazso: It's a bit confusing that we also use our type representing store data `AppState`.

src/actions/asyncActions.ts
```diff
@@ -0,0 +1,460 @@
...
+    RecentPostFeed,
+} from '../social/api';
+import { Post } from '../models/Post';
+import { ImageData } from '../models/ImageData';
+import { isContactFeed, makeContactFromRecentPostFeed } from '../helpers/feedHelpers';
+import { ContactFeed } from '../models/ContactFeed';
+import { Contact } from '../models/Contact';
+import { ContactActions } from './ContactActions';
+
+type Thunk = (dispatch: any, getState: () => AppState) => Promise<void>;
```
agazso: These definitions are also in `actionHelpers.ts`

src/actions/asyncActions.ts
```diff
@@ -0,0 +1,460 @@
...
+type ThunkTypes = Thunk | Actions;
+
+const isActionTypes = (t: ThunkTypes): t is Actions => {
```
agazso: This could go to `actionHelpers.ts` too

src/store/legacy.ts
```diff
@@ -0,0 +1,56 @@
+import { PersistConfig, createMigrate, KEY_PREFIX, getStoredState } from 'redux-persist';
```
agazso: Maybe a better name for this file would be `legacyStorage.ts` or `legacyPersistConfig.ts`

vujevits: Renamed it to `legacyAsyncStorage`

agazso:
After testing it on my machine, I see the following issues:

On both platforms:
- [ ] After installing the new version there is an error at startup and the app is loaded with data from the old version. I think saving in general is not working anymore. This may be the root cause for other issues as well:
https://gist.github.com/agazso/a3ca29ebc29657457bfa280b79b489cb

On iOS:
- [ ] The share extension does not appear in the browser's share menu, even if I click on `More...`. First  you have to enable it from the Felfele app's share menu, clicking on `More...` and enabling it from the list of activities
- [ ] Then sharing from the browser opens the share extension and the post editor, but when clicking on the Send button it changes to a spinner and nothing happens.
- [ ] The `AppRegistry.registerComponent` call is missing from `index.ios.js`. In general that file should be identical to `index.android.js`.

On Android:
- [ ] The post editor disappears, but the post does not appear in the app. This may be related to the `redux-persist` issue

Other issues (may be fixed in a different PR):
- [ ] When sharing a photo a new post is created with a `file://` link as text. The app should detect supported files (right now it's images only) and share them correctly and ignore everything else. Maybe it's possible to filter at the system-wide Share dialog level.

vujevits:
- Not sure what do you mean by your first point in iOS, where do you have to enable it? It should be listed in "More"
- the `index.ios.js` and `index.android.js` are different with a reason, actually when they were not different, and there should have been just an `index.js`, but there is a known bug which sometimes prevents this, but it has low priority in react-native's agenda. For iOS there is a `share.index.js`, which serves as an entry point for the extension 

I'll look at the rest of the issues, there was lot of back and forth working with different versions, and especially the Android simulator was quite challenging to work with.

agazso:
Ok, the migration seems to be working so the data is stored correctly. 

Things that I discovered during debugging:
- It's possible to see the logs from the Simulator in the Console app on macOS
- The share extension does not appear in Safari in the Simulator for some reason, you have to activate it in an other app first
- Sharing does not seem to be working in the Simulator, only when starting from XCode with a specific app

Issues remaining:

On iOS:
- [ ] Sharing a photo from the Photos app crashes the share extension, here is the crash log:
https://gist.github.com/agazso/843f28adf2ed0f26495f157f3c87cd43

On Android:
- [ ] When sharing a photo a new post is created with a file:// link as text. The app should detect supported files (right now it's images only) and share them correctly and ignore everything else.

vujevits:
I disabled sharing images for now, so it can be addressed in a separate PR.

vujevits:
opened an issue for image sharing https://github.com/felfele/felfele/issues/448

Issue #14, opened 2018-10-16
Messaging based on mutual follow


vujevits:
You can send messages to people with whom you are mutually follow each other. You will get notified if someone sends you a message.

PR #447, opened 2019-08-02
WIP Private group messaging protocol

agazso:
Implements the protocol for #15.

- Private groups are defined by their participant list and a shared secret
- Anyone in the group can invite more participants
- Uses `Timeline` and `Chapter` types from [Erebos](https://github.com/MainframeHQ/erebos)
- There are one private and one group channel for all the participants, both are implemented as `Timeline`s
- Participants can only be added, they cannot be removed
- Right now the shared secret cannot be changed (no forward and backward secrecy)


agazso:
There is a new `Storage` type which represents the persistent storage that can be used in different protocols. At the moment there are two implementations:
- `SwarmStorage` uses Swarm as storage
- `MemoryStorage` which is not really persistent, because it stores data in memory, can be used for testing

With this it's possible to create protocol code that can be tested locally and when it works it can be tested on Swarm as well easily. I expect that after this iterating on the protocols becomes quicker.

vujevits:
> There is a new `Storage` type which represents the persistent storage that can be used in different protocols. At the moment there are two implementations:
> 
>     * `SwarmStorage` uses Swarm as storage
> 
>     * `MemoryStorage` which is not really persistent, because it stores data in memory, can be used for testing
> 
> 
> With this it's possible to create protocol code that can be tested locally and when it works it can be tested on Swarm as well easily. I expect that after this iterating on the protocols becomes quicker.

Can we have a more descriptive name for `Storage`?

vujevits:
> Implements the protocol for #15.
> 
> Changes proposed in this pull request:
> 
>     * Private groups are defined by their participant list and a shared secret
> 
>     * Anyone in the group can invite more participants
> 
>     * Uses `Timeline` and `Chapter` types from [Erebos](https://github.com/MainframeHQ/erebos)
> 
>     * There are one private and one group channel for all the participants, both are implemented as `Timeline`s
> 
>     * Participants can only be added, they cannot be removed
> 
>     * Right now the shared secret cannot be changed (no forward and backward secrecy)

Not sure I understand what the 4th point is referring to

vujevits:
Also #14

agazso:
> Can we have a more descriptive name for `Storage`?

Renamed to `ProtocolStorage`

agazso:
So it turns out we need to have two distinct protocols:
- One for private sharing with a contact, encrypted with their derived shared key
- An other for sharing with a group. This requires the first one to work correctly, because invitations to the group happen outside the group privately.

This PR implements the first one in a way that can be used later for the second. There are two Timelines for each private channel. Each participant can write their own and the other can download it, then the protocol generates a list of `Post`s on from them that is identical for both parties.

There is also code so that it's possible to write test cases for the protocol in a declarative way. It turns out that it requires significant amount of scaffolding so that any stateful component can be injected in a code that deals with networking, crypto, random and time for multiple participants at the same time.

With this all the private sharing protocol tests can be run from Jest, CLI and it's possible to run them on Swarm as integration test, however as it is stateful it has to be stopped and the data files should be deleted between running tests.

This PR does **not** implement the group protocol (#15) yet, only the private sharing protocol (#14).

agazso:
> I assume we are not using `erebos` types from its source because it is entangled with its Bzz implementation?

Yes, I did not want to deal with issues from the `erebos` package in this PR, but I copied the type definitions from their codebase.

PR #464, opened 2019-08-26
Private Channels view

### LINKED ISSUE #456: Private channels tab instead of Favorites
The second tab should show Private channels instead of Favorites.

https://zpl.io/V19YOrm

All Private channels should show the contacts

https://zpl.io/b6gY49j

Private channels should show only the posts from you and the other participant

https://zpl.io/25A3Jd8

On the Info page you can delete the contact

https://zpl.io/boLr8Ov

agazso:
Fixes #456 .


agazso:
This PR adds the `Private channels` tab, the `All private channels` and `Private channel page` screens from the current design.

All references to Favorites are removed.

src/actions/asyncActions.ts
```diff
@@ -119,6 +126,58 @@ export const AsyncActions = {
...
+                                uri: contactFeed.feedUrl,
+                                image: contact.image,
+                            },
+                        };
+                        allPosts.push(postWithAuthor);
+                    }
+                }
+            }
+            const posts = mergeUpdatedPosts(allPosts, previousPosts);
+            dispatch(Actions.updateRssPosts(posts));
```
vujevits: This action should be renamed

src/actions/asyncActions.ts
```diff
@@ -119,6 +126,58 @@ export const AsyncActions = {
             dispatch(Actions.updateRssPosts(posts));
         };
     },
+    downloadPrivatePostsFromContacts: (contactFeeds: ContactFeed[]): Thunk => {
```
vujevits: Is there a way to refresh when you are on a view with a specific 1to1 feed?

agazso: You can pass in an array with one element.

src/actions/asyncActions.ts
agazso: This updates the `rssPosts` field in the AppState, hence the name.

What is your suggestion?

PR #468, opened 2019-08-28
fix the invite link being converted to inviteCode too soon

vujevits:
Fixes broken contact invite


src/components/FeedInfo.tsx
```diff
@@ -317,13 +317,11 @@ export class FeedInfo extends React.Component<Props, FeedInfoState> {
         try {
             Debug.log('FeedInfo.onScanSuccess', 'data', data);
             const inviteLink = getInviteCodeFromInviteLink(data);
```
agazso: This line seems unnecessary after the change

PR #473, opened 2019-09-02
implement screens and navigation for contact flow

### LINKED ISSUE #461: New contact flow

vujevits:
Fixes #461 
- confirm and success screen for receiver
- success screen for sender

Another PR will take care of better polling and what to do while waiting.


src/styles.ts
```diff
@@ -30,6 +32,7 @@ export const ComponentColors = {
     BACKGROUND_COLOR: Colors.LIGHT_GRAY,
     DISABLED_BUTTON_COLOR: Colors.LIGHT_GRAY,
     BUTTON_COLOR: Colors.BRAND_PURPLE,
+    SECONDARY_RECT_BUTTON_COLOR: Colors.LIGHTER_GRAY_2,
```
agazso: What does RECT mean here?

src/styles.ts
```diff
@@ -14,8 +14,10 @@ export const Colors = {
     BROWNISH_GRAY: '#696969',
     LIGHT_GRAY: '#DDDDDD',
     LIGHTER_GRAY: '#E6E6E6',
+    LIGHTER_GRAY_2: '#EEEEEE',
```
agazso: I also introduced this, but I just replaced `LIGHTER_GRAY` with this color.

src/styles.ts
vujevits: Rectangular

src/styles.ts
vujevits: great! I was running out of names 😄

src/styles.ts
agazso: Can it be simply `SECONDARY_BUTTON_COLOR`? With this I get the impression that there should be a `PRIMARY_RECT_BUTTON_COLOR` as well (and in that case I wouldn't shorten rectangular).

vujevits: that would be misleading, because, there is a `BUTTON_COLOR`, and it is not the corresponding secondary option color. I added a primary one and remove shortening.

PR #497, opened 2019-09-11
advance contacts state on refresh; removed confirm from mutual contact

### LINKED ISSUE #491: contact state should be updated at app restart and with manual trigger as well

### LINKED ISSUE #492: Name of the contact should be passed along with the link
This is needed for displaying a confirmation dialog at the beginning of the contact flow, which also works when users are offline

vujevits:
Fixes #491
Fixes #492

- [x] as a short term solution name of the sender is sent in the link, no image
- [x] pull to refresh on private channels and app start triggers contact state advancing
- [x] confirmation is not stored, it is just a popup at the beginning
- [x] remove legacy functionality from feedInfo: follow link, contact feed


src/helpers/feedHelpers.ts
```diff
@@ -110,3 +109,17 @@ export const fetchFeedFromUrl = async (url: string, swarmGateway: string): Promi
         return null;
     }
 };
+
+export const fetchRSSFeedFromUrl = async (url: string): Promise<Feed | null> => {
+    try {
+        Debug.log('fetchFeedFromUrl', 'url', url);
```
agazso: The name in the `Debug.log` does not match the name of the function

agazso:
👍

PR #501, opened 2019-09-16
Changed hitSlop to zero in case of GridCards

agazso:
- TouchableView has a default hitSlop, it is turned of in case of GridCards


vujevits:
👍

PR #505, opened 2019-09-17
New contact tab

### LINKED ISSUE #458: New Contact tab
Instead of the Profile page there should be a Contact page, with QR code, Scan and Share buttons and a link to the Profile editor page.

https://zpl.io/2jYB51x

https://zpl.io/2jYB51q

agazso:
Fixes #458  .

- Renamed IdentitySettings to ContactScreens
- Added EditProfile screen
- All of these are in `ui/screens/profile`

Should be merged after #503


src/selectors/selectors.ts
```diff
@@ -135,6 +137,28 @@ export const getPrivateChannelPosts = createSelector([ getSelectedTopicPosts ],
     return posts;
 });
 
-export const getYourPosts = createSelector([ getLocalPosts, getProfile ], (posts, author) => {
-    return posts.filter(post => post.author && post.author.name === author.name);
+export const getYourPosts = createSelector([ getLocalPosts, getPrivatePosts, getProfile ], (posts: Post[], privatePosts: PostListDict, author) => {
+    const yourPubicKey = author.identity!.publicKey;
```
vujevits: 18+ naming convention 👀

src/selectors/selectors.ts
```diff
@@ -135,6 +137,28 @@ export const getPrivateChannelPosts = createSelector([ getSelectedTopicPosts ],
...
+    const yourPrivatePosts = Object
+        .values(privatePosts)
+        .reduce((prev, curr) => prev.concat(curr), [])
```
vujevits: It's not clear to me what are you trying to achieve here. It seems like Object.values returns an array, and then in the `reduce` you clone the array. Is this handling some kind of an edge case?

src/ui/screens/profile/EditProfile.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    name === ''
+                    ? props.onUpdateAuthor(DEFAULT_AUTHOR_NAME)
+                    : props.onUpdateAuthor(name)
+                }
+            />
+        </ScrollView>
+    </FragmentSafeAreaViewWithoutTabBar>
+);
+
+const WIDTH = Dimensions.get('window').width;
```
vujevits: it's advised to always get this value in the `render` function

src/selectors/selectors.ts
agazso: `privatePosts` is an Object, the property names are the topics, and their values are list of posts.
Therefore `Object.values` returns a list of lists and the reduce is flatting it out to a single list.

src/ui/screens/profile/EditProfile.tsx
agazso: I read about this too, the main reason is when you support both portrait and landscape and you want to keep your app adaptive.

Here we use it for calculating style properties so those calculations should go to the render function. As we don't support landscape I didn't bother, also most of this component is a copy of the ProfileScreen from the onboarding.

src/selectors/selectors.ts
vujevits: I see. I find this alternative easier to read:
```js
const postsByTopics = Object.values(privatePosts);
const yourPrivatePosts = [].concat(...postsByTopics);
```

agazso: I changed the reduce to concat.

Unfortunately Typescript cannot infer the types correctly so you have to explicitly write out the types.

PR #549, opened 2019-09-27
Expire invite

### LINKED ISSUE #494: Expire contact links
Contact links should expire both for security and optimization.

vujevits:
Fixes #494


src/ui/screens/contact/ContactConfirm.tsx
```diff
@@ -60,7 +61,8 @@ export const ContactConfirm = (props: Props) => {
                     />
                 </View>
                 <RegularText style={{ textAlign: 'center', color: Colors.BROWNISH_GRAY }}>
-                    {`Connect with ${props.inviteCode.profileName}?`}
+                    {`Connect with ${props.inviteCode.profileName}?
+This link will expire in ${DateUtils.printableElapsedTime(props.inviteCode.expiry!)}`}
```
agazso: It seems that the exclamation mark after `expiry` is not necessary

src/models/InviteCode.ts
```diff
@@ -1,7 +1,27 @@
...
-    contactPublicKey: HexString;
-    profileName?: string;
+    [InviteCodeFields.version]: number;
+    [InviteCodeFields.randomSeed]: HexString;
+    [InviteCodeFields.contactPublicKey]: HexString;
+    [InviteCodeFields.profileName]: string;
+    [InviteCodeFields.expiry]: number;
 }
+
+export const isVersion1RawInviteCode = (jsonObj: any): boolean => {
```
vujevits: @agazso there will be a better solution for this in TypeScript 3.7, called `assert signatures`. If you check the call site you can see that it would make sense to throw the error from such function. 
https://httptoolkit.tech/blog/5-big-features-of-typescript-3.7/

src/helpers/deepLinking.ts
```diff
@@ -20,21 +20,35 @@ const urlSafeBase64Decode = (data: string) => {
...
 
 export const getInviteLink = (contact: InvitedContact, profileName: string): string => {
+    return `${BASE_URL}${INVITE}${makeInviteParams(contact, profileName)}`;
+};
+
+export const getInviteLinkWithParams = (params: string): string => {
+    return `${BASE_URL}${INVITE}${params}`;
+};
+
+const makeInviteParams = (contact: InvitedContact, profileName: string): string => {
```
vujevits: @agazso tbh it doesn't feel right to have the parameters with different encoding (base64, raw string, urlEncoded), wdyt?

agazso: I don't think it's important, because the links are meant to be copy-pasted and not for human input.

However it's more important that the links should be short, because the longer they are the harder to scan the QR codes that are generated from them, especially under not ideal light conditions. Also long links in messenger apps look messy and are hard to use, so basically it's a UX issue.

That was the original reason that the `randomSeedBytes` and `contactPublicKeys` were base64 encoded, because they contain 32 bytes binary data. If we used hexstrings they would have been 64 bytes each but with base64 encoding they can be stored in 44 bytes.

agazso:
I would do the versioning differently for two reasons:
- As I mentioned in my comment regarding the encoding, it's better UX if the links are shorter, so I would not encode the name of the fields in the link, because by that we only gain the flexibility of changing the ordering of the fields, but that's not really important. In turn it makes the links 25-30% longer.
- The suggested solution in this PR introduces a lot of complexity (const enum, json conversion) while it's not really clear how it would solve the original versioning problem when the `version` changes in the invite code.

Instead I suggest a more simple solution where the `params` of the invite link encodes fields that are separated with the `SEPARATOR`. The first field is always the version and based on that it is known what are the order of the other fields.

```typescript
const [version, ...rest] = inviteLink.split(SEPARATOR);
switch (version) {
    case 1: return parseVersion1Params(rest);
    default: throw new Error('unknown version');
}

// ...

const parseVersion1Params = (params: string[]): InviteCode => {
    if (params.length != 4) {
        throw new Error('invalid parameters');
    }
    // etc.
}

```

src/models/InviteCode.ts
```diff
@@ -1,7 +1,27 @@
 import { HexString } from '../helpers/opaqueTypes';
 
+export const enum InviteCodeFields {
```
agazso: I get errors like this from the Metro bunder on this line: 
`'const' enums are not supported.`

Interestingly the `npm run check` does not report an error on this.

vujevits: yes, it is caused by babel

vujevits:
thanks @agazso , I see now what you meant with the versioning previously 👍

src/helpers/deepLinking.ts
```diff
@@ -38,11 +37,11 @@ const makeInviteParams = (contact: InvitedContact, profileName: string): string
...
     return `\
-${InviteCodeFields.randomSeed}=${base64RandomSeed}${SEPARATOR}\
-${InviteCodeFields.contactPublicKey}=${base64ContactPublicKey}${SEPARATOR}\
-${InviteCodeFields.profileName}=${encodeURIComponent(profileName)}${SEPARATOR}\
-${InviteCodeFields.expiry}=${contact.createdAt + CONTACT_EXPIRY_THRESHOLD}${SEPARATOR}\
-${InviteCodeFields.version}=${INVITE_CODE_VERSION}\
+${INVITE_CODE_VERSION}${SEPARATOR}\
+${base64RandomSeed}${SEPARATOR}\
+${base64ContactPublicKey}${SEPARATOR}\
+${contact.createdAt + CONTACT_EXPIRY_THRESHOLD}\
```
agazso: `SEPARATOR` is missing here.

agazso:
I also suggest to make a test for at least a correct and an incorrect link.

vujevits:
good point!

test/helpers/deepLinkingTest.ts
```diff
@@ -25,3 +25,18 @@ test('invite link encoding and decoding', () => {
...
+    const inviteLink = 'https://app.felfele.org/invite/1&7NWsGi+_Zad6oxtWcZnG6a3IDIg96aXzI59rGK08No8=&A0htzKYoCkSsjKbKMWhGTxYpWkSlf+uDiUILUXLbRLXO&1570552184258&fff';
+    expect(() => getInviteCodeFromInviteLink(inviteLink)).not.toThrow();
+});
+
+test('invite link with unknown version throws error', () => {
+    const inviteLink = 'https://app.felfele.org/invite/2&7NWsGi+_Zad6oxtWcZnG6a3IDIg96aXzI59rGK08No8=&A0htzKYoCkSsjKbKMWhGTxYpWkSlf+uDiUILUXLbRLXO&1570552184258&fff';
+    expect(() => getInviteCodeFromInviteLink(inviteLink)).toThrow('unknown version');
+});
+
+test('version 1 malformed invite link throws error', () => {
```
agazso: This is the same test as the previous

src/helpers/deepLinking.ts
```diff
@@ -20,41 +24,50 @@ const urlSafeBase64Decode = (data: string) => {
...
     const randomSeedBytes = hexToUint8Array(contact.randomSeed);
-    const contactPulicKeyBytes = hexToUint8Array(contact.contactIdentity.publicKey);
+    const contactPublicKeyBytes = hexToUint8Array(contact.contactIdentity.publicKey);
     const base64RandomSeed = urlSafeBase64Encode(randomSeedBytes);
-    const base64ContactPublicKey = urlSafeBase64Encode(contactPulicKeyBytes);
-    const inviteLinkWithoutUsername = getInviteLinkWithBase64Params(base64RandomSeed, base64ContactPublicKey);
-    return `${inviteLinkWithoutUsername}/${encodeURIComponent(profileName)}`;
+    const base64ContactPublicKey = urlSafeBase64Encode(contactPublicKeyBytes);
+    return `\
+${INVITE_CODE_VERSION}${SEPARATOR}\
```
agazso: It would be less error prone to pass the values in an array and invoke `.join(SEPARATOR)` on it.

src/helpers/deepLinking.ts
```diff
@@ -20,41 +24,50 @@ const urlSafeBase64Decode = (data: string) => {
...
+${base64RandomSeed}${SEPARATOR}\
+${base64ContactPublicKey}${SEPARATOR}\
+${contact.createdAt + CONTACT_EXPIRY_THRESHOLD}${SEPARATOR}\
```
agazso: Dates in Javascript are stored as milliseconds, however we don't really need that kind of precision. I think it would be better to store date in seconds and it would make the link a bit shorter too.
In that case the parsing code needs to be changed too.

PR #553, opened 2019-10-02
improve post flow

vujevits:
- [x] if only one contact is selected when sharing, the app should navigate to the appropriate feed
- [x] if someone starts post editing from a private channel, it should be shared without the share dialog

### DIFF
--- a/src/ui/screens/contact/ContactGrid.tsx
+++ b/src/ui/screens/contact/ContactGrid.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { StyleSheet, View, SafeAreaView } from 'react-native';
+import { StyleSheet, View } from 'react-native';
 import { SuperGridSectionList } from 'react-native-super-grid';
 
 import { ContactFeed } from '../../../models/ContactFeed';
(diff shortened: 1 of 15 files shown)


src/ui/screens/contact/ContactGrid.tsx
```diff
@@ -10,14 +10,15 @@ import { getFeedImage } from '../../../helpers/feedHelpers';
...
 
 export interface DispatchProps {
-    onPressFeed: (feed: ContactFeed) => void;
+    onPressFeed: (feed: ContactFeed | Feed) => void;
 }
 
 export interface FeedSection {
     title?: string;
-    data: ContactFeed[];
+    data: Array<Feed | ContactFeed>;
```
agazso: Why this definition has changed? On the ContactGrid we can only display `ContactFeed`s, not ordinary `Feed`s.

src/ui/screens/private-channels/PrivateChannelsListView.tsx
```diff
@@ -7,15 +7,16 @@ import { ContactGrid } from '../contact/ContactGrid';
...
 export interface DispatchProps {
-    onPressChannel: (feed: ContactFeed) => void;
+    onPressChannel: (feed: ContactFeed | Feed) => void;
     onAddChannel: () => void;
 }
 
 export interface FeedSection {
     title?: string;
-    data: ContactFeed[];
+    data: Array<Feed | ContactFeed>;
```
agazso: Why this definition has changed? On the PrivateChannelListView we can only display ContactFeeds, not ordinary Feeds.

vujevits: I was unsure about both of these occurrences because of the sharing with self usecase

PR #556, opened 2019-10-02
iOS beta release

### LINKED ISSUE #554: Beta or staging release
It would be good to have a beta or staging release which could be a completely different app that could be installed on the phone besides the release version of the app.

Then it could have its own data stored that could be used for testing and it could be safely deleted or reinstalled any time, keeping the data of the release version safe.

agazso:
Fixes #554 on ios .

- Added Beta configuration to XCode project
- Workaround for [React Native xcode configuration issue](https://github.com/facebook/react-native/issues/11813) with [React Native Schemes Manager](https://github.com/thekevinbrown/react-native-schemes-manager) 
- Increased build number


package.json
```diff
@@ -34,6 +34,7 @@
     "node-fetch": "2.3.0",
     "react-native-asset": "1.1.2",
     "react-native-debugger-open": "0.3.17",
+    "react-native-schemes-manager": "^2.0.0",
```
vujevits: This one is not pinned

agazso: 👍

